PR #2, opened 2019-01-28
Add a simple static file server

honza:
It serves the assets from `build/` in preparation for the `create-react-app` app.


honza:
This is how the [dep](https://golang.github.io/dep/) tool works. See this example from [openshift installer](https://github.com/openshift/installer/tree/master/vendor)

honza:
There is also this [FAQ](https://golang.github.io/dep/docs/FAQ.html#should-i-commit-my-vendor-directory) in dep's docs.

flofuchs:
> There is also this [FAQ](https://golang.github.io/dep/docs/FAQ.html#should-i-commit-my-vendor-directory) in dep's docs.

Thanks Honza!

PR #3, opened 2019-01-29
Initial react app

jtomasek:


jtomasek:
We need to figure out correct names usage in package.json, index.html and manifest.json. Suggestions welcome! README needs to get updated and for some reason, background image does not work. I think these things can get updated in subsequent PRs.

src/App.tsx
```diff
@@ -1,26 +1,164 @@
...
-            className="App-link"
-            href="https://reactjs.org"
-            target="_blank"
-            rel="noopener noreferrer"
-          >
-            Learn React
-          </a>
-        </header>
-      </div>
+      <Fragment>
```
jtomasek: Note that the text and inline styles are just POC of desired layout.

honza:
:shipit:

PR #6, opened 2019-02-01
Add redux

honza:


jtomasek:
I am not exactly sure if we want to have a router. I think we always want the rendered screen to be determined strictly by application state. Since the application is just a wizard, we won't make use of browser history buttons or let user insert a url path to access certain view or even use `<Link>`s.

Router seems to complicate things for us as well, e.g. user enters /nodes path, Since the router routes define which component is rendered, Nodes components gets rendered. Then Nodes component needs to look at App state to determine if it is supposed to redirect to another route. Redirect happens and correct component gets rendered.

Without router it is just: root component looks at app state and renders the component which is determined by app state.

honza:
I'm still going to try and write a connected component to see if it's all hooked up properly.

honza:
Ready for review, I think

src/index.tsx
```diff
@@ -1,10 +1,18 @@
...
 import ReactDOM from 'react-dom';
+import { Provider } from 'react-redux';
 import './styles/index.scss';
 import App from './App';
 import * as serviceWorker from './serviceWorker';
+import { store } from './store';
 
-ReactDOM.render(<App />, document.getElementById('root'));
+const Root = () => (
+  <Provider store={store}>
```
jtomasek: How about keeping the index.tsx intact and do this in App component instead?

jtomasek:
yarn.lock update is missing. We should experiment with colocating actions, reducers, selectors... to the components in files structure. Let's keep it the way it is now though.

src/index.tsx
honza: Done

honza:
* Move `Provider` to `App`
* Rename `root-reducer` to `rootReducer`
* Rename `GET_NODES` constant from `get-nodes` to `getNodes`
* Add missing return types

PR #9, opened 2019-02-04
Add a simple API handler

honza:
During development, you can take advantage of using the Golang server.  The server does two things:

* It provides a REST API layer (e.g. `/api/nodes`)
* It reverse-proxies requests to the `react-scripts` development server

To use it, start the `yarn start` server in one tab, and then start the Golang server with `go run main.go server --use-proxy` in another tag.  This way you can access both from a single hostname and port.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,11 +2,37 @@
 
 User interface for MetalKube
 
+## Getting started
+
+* Install [yarn][1] and [golang][2]
+* Set up your `$GOPATH` (this means selecting a directory on your computer where
+  all your golang code will go, and setting the `GOPATH` environment variable to
+  that path)
+* Clone this repository to your `$GOPATH` directory, i.e.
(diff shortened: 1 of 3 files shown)


honza:
If the golang server will be serving the app in production, I think it makes more sense to serve the app in development, too. But I guess nothing is preventing you from doing it the way you like it better.

honza:
Let me rename the nodes to hosts.

honza:
* Renamed nodes to hosts
* Added documentation to the README for setting up the project

honza:
@jtomasek OK, I found a reason why my proxy direction won't work during development: the go server doesn't proxy websocket connections, and those are needed for rebuilds and reloads. I'm not interested in implementing that and maintaining that. I'll update this PR to remove the proxy.

honza:
* Removed proxy server

honza:
* Rebased on top of `master`

honza:
* Another rebase on `master`

PR #12, opened 2019-02-11
Add PatternFly3 and patternfly-react

jtomasek:
This allows the app to use patternfly-react components (PatternFly v3 based) when needed.


jtomasek:
> So now we'll have patternfly 3 and -next ?

Right. It aligns us to the current status of openshift/console. PF3 is just in case fallback, but it is important to make sure we can use it if needed.

PR #14, opened 2019-02-11
Add Redux infrastructure for hosts

honza:
This PR adds the project structure for interacting with the REST API.

* redux actions
* redux reducer
* api layer with custom API response types
* an example hosts component which is connected to redux

This is rebased on top of #9


honza:
* Rebase on top of master now that #9 is merged

honza:
* Rebase on top of master

honza:
`typesafe-actions` doesn't seem to export the return types for `action()` so it's kind of tricky; but I'll have another look

src/actions/hosts.ts
```diff
@@ -0,0 +1,33 @@
...
+
+interface HostListApiResponse {
+  data: Host[];
+}
+
+export const getHosts = () => action(GET_HOSTS_REQUEST);
+export const getHostsSuccess = (data: HostListApiResponse) =>
+  action(GET_HOSTS_SUCCESS, data);
+
+export const getHostsFailure = () => action(GET_HOSTS_FAILURE);
```
jtomasek: IIUC, these 3 action definitions could get simplified as:
```
const fetchHosts = createAsyncAction(
  GET_HOSTS_REQUEST,
  'GET_HOSTS_SUCCESS',
  'GET_HOSTS_FAILURE'
)<void, Host[], Error>();
```
then
```
const fetchHostsAsync = () => (dispatch: redux.Dispatch) => {
  dispatch(fetchHosts.request());
  getHosts()
    .then(response => {
      dispatch(fetchHosts.success((response.data)));
    })
    .catch(() => {
      dispatch(fetchHosts.failure(Error('some error)));
    });
};
```

I find this better because of naming. In api modules, we could use HTTP verbs (getHosts, postHosts, getHost...) and selectively import it here rather than using *.

src/api/hosts.ts
```diff
@@ -0,0 +1,9 @@
+import axios, { AxiosPromise } from 'axios';
+import { Host } from '../models/hosts';
+import { ApiResponse } from './index';
+
+type HostListApiReponse = AxiosPromise<ApiResponse<Host[]>>;
+
+// TODO: Add configuration for the API endpoint
+export const getHosts = (): HostListApiReponse =>
+  axios.get('http://localhost:8080/api/hosts');
```
jtomasek: `axios.get('/api/hosts');` now works too

src/actions/hosts.ts
honza: Man, I remember looking at those examples, and it wasn't clicking for me. I get it now, and yes, this is good.

src/api/hosts.ts
honza: :sparkles:

src/api/hosts.ts
```diff
@@ -0,0 +1,7 @@
+import axios, { AxiosPromise } from 'axios';
+import { Host } from '../models/hosts';
+import { ApiResponse } from './index';
+
+type HostListApiReponse = AxiosPromise<ApiResponse<Host[]>>;
```
jtomasek: Nit: I am inclined to rename this to 'GetHostsApiResponse'

honza: Yep, works for me.

PR #11, opened 2019-02-06
Add a production build script

honza:
This is branched from #9. Once that PR is merged, this will need to be rebased on top of master.

This adds a script to produce a production binary for the facet server.

The README now contains documentation for this process.

PR #15, opened 2019-02-11
Long running tasks

honza:
See the README for more info.

To test, start the server, and open the app. Then, in another browser window or with curl, you can hit the http://localhost:8080/api/long endpoint. You should see messages printed out from the server process, and in the developer console in the browser.

Rebased on top of #11


knowncitizen:
I don't think we want vendor binaries?

honza:
* Rebased again on top of #11

honza:
* Rebased on top of `master` now that #11 is merged

README.md
```diff
@@ -70,6 +70,47 @@ Use "facet [command] --help" for more information about a command.
 
 [3]: https://github.com/rakyll/statik
 
+## Long-running tasks
+
+If you would like to trigger a long-running task from the frontend, we've got
```
knowncitizen: Neat.

honza:
> Where does log go? Is it configurable?

By default, it's just printed to stdout from the server process. Yes, it can be configured --- what did you have in mind?

knowncitizen:
Was just curious. Thanks for the insight.

honza:
* Rebase on master

pkg/server/routes.go
```diff
@@ -34,11 +35,45 @@ type ApiResponse struct {
...
+	}
+
+	return values[ns]
+}
+
+func (ns NotificationStatus) MarshalJSON() ([]byte, error) {
+	return json.Marshal(ns.String())
+}
+
+type Notification struct {
```
jtomasek: At some point, we'll need to figure out standardized notification format, probably aligned to day2 (I expect 'type' to identify the message and 'payload' for additional data)

src/api/websocket.ts
```diff
@@ -0,0 +1,22 @@
+// TODO: This is just a dumb client to test the websocket connection
+// TODO: This needs to be connected to the redux store
```
jtomasek: We can experiment with putting this into a React Component which would handle websocket errors, reconnecting and dispatching the actions based on notification type

PR #21, opened 2019-02-13
Add CreateClusterForm

jtomasek:
Introduce basic form fields handling
Add @patternfly/react-icons to allow for using icons

PR #22, opened 2019-02-14
Cluster wizard steps

jtomasek:
Depends on #21


src/components/ClusterWizard.tsx
```diff
@@ -1,18 +1,19 @@
...
 
-// import BaremetalInventory from './BaremetalInventory';
+import BaremetalInventory from './BaremetalInventory';
 import ClusterWizardNav from './ClusterWizardNav';
-// import ClusterWizardSteps from './ClusterWizardSteps';
 import CreateClusterForm from './createCluster/CreateClusterForm';
 
-const ClusterWizard: FC = (): JSX.Element => (
+interface ClusterWizardProps {
+  currentStepIndex: number;
```
honza: Can we still use a more concrete type to represent the wizard step?

E.g.:

```tsx
enum WizardStep {
    One,
    Two
}

let step: WizardStep = WizardStep.One;
```

It gives you much nicer comparisons:

```tsx
if (step === WizardStep.One) {
    
}
```

Instead of comparing against numbers.

jtomasek: Done

lizsurette:
@jtomasek Looking great so far! Hopefully the PatternFly HTML/CSS will land soon to pull from, I'm just noticing small tweaks needed like a bubble behind the numbers...
screen shot 2019-02-15 at 10 54 21 pm
screen shot 2019-02-15 at 10 54 06 pm

PR #25, opened 2019-02-14
Add $GOPATH/bin for statik invocation

dantrainor:


hardys:
Ah good point, I didn't spot this when we removed the $PATH addition in dev-scripts #66

PR #19, opened 2019-02-13
Add an API endpoint for booting the bootstrap VM

honza:
You are expected to run the facet server on the dev-scripts host.

There is a [PR](https://github.com/metalkube/dev-scripts/pull/66) for dev-scripts that installs facet.

Depends on #30

PR #26, opened 2019-02-14
API endpoint for creating ignition configs

honza:
This uses openshift/installer as a library.

Please note the new instructions on how to run the server.

### DIFF
--- a/pkg/server/routes.go
+++ b/pkg/server/routes.go
@@ -29,6 +29,7 @@ func ApiRouter(router *mux.Router, notificationChannel chan common.Notification)
 	router.HandleFunc("/long", LongRunningTaskHandler(notificationChannel))
 	router.HandleFunc("/bootstrap-vm", BootstrapVMHandler(notificationChannel))
 	router.HandleFunc("/cluster-definition", CreateClusterDefinition).Methods("POST")
+	router.HandleFunc("/create-ignition-configs", CreateIgnitionConfigsHandler)
 }
 
 func CreateRouter(notificationChannel chan common.Notification, websocketWorker *WebsocketWorker) *mux.Router {
--- a/pkg/server/routes.go
+++ b/pkg/server/routes.go
@@ -29,7 +29,7 @@ func ApiRouter(router *mux.Router, notificationChannel chan common.Notification)
 	router.HandleFunc("/long", LongRunningTaskHandler(notificationChannel))
 	router.HandleFunc("/bootstrap-vm", BootstrapVMHandler(notificationChannel))
 	router.HandleFunc("/cluster-definition", CreateClusterDefinition).Methods("POST")
-	router.HandleFunc("/create-ignition-configs", CreateIgnitionConfigsHandler)
+	router.HandleFunc("/ignition-configs", CreateIgnitionConfigsHandler)
 }
 
 func CreateRouter(notificationChannel chan common.Notification, websocketWorker *WebsocketWorker) *mux.Router {
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ For further details about the MetalKube architecture, see
 ## Running the production server
 
 ```
-$ go run main.go server
+$ go run -tags "libvirt release" main.go server
 ```
 
 ## Running the development server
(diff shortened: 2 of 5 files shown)


honza:
* Rebase on #19

honza:
* Rebase on top of master

honza:
* Another rebase on master

pkg/server/routes.go
```diff
@@ -28,6 +28,7 @@ func ApiRouter(router *mux.Router, notificationChannel chan common.Notification)
 	router.HandleFunc("/hosts", HostsHandler)
 	router.HandleFunc("/long", LongRunningTaskHandler(notificationChannel))
 	router.HandleFunc("/bootstrap-vm", BootstrapVMHandler(notificationChannel))
+	router.HandleFunc("/create-ignition-configs", CreateIgnitionConfigsHandler)
```
hardys: Having "create" in the path doesn't seem all that ReSTful, should we perhaps just GET api/ignition/configs, or optionally POST to ignition-configs with an install-config.yaml as the payload?

hardys:
For development this will be useful, but I kinda wonder if we'll ultimately combine bootstrap-vm and create-ignition-configs, I guess we can iterate on the individual pieces then combine them later though

pkg/server/routes.go
honza: Yes, good point.

honza: Done

pkg/server/routes.go
```diff
@@ -28,6 +28,7 @@ func ApiRouter(router *mux.Router, notificationChannel chan common.Notification)
 	router.HandleFunc("/hosts", HostsHandler)
 	router.HandleFunc("/long", LongRunningTaskHandler(notificationChannel))
 	router.HandleFunc("/bootstrap-vm", BootstrapVMHandler(notificationChannel))
+	router.HandleFunc("/ignition-configs", CreateIgnitionConfigsHandler)
```
jtomasek: I like this way of defining routes. It would be nice to establish a standardized way of defining routes and handlers e.g.:
```
router.HandleFunc("/hosts", GetHostsHandler).Methods("GET")
router.HandleFunc("/hosts/{id}", GetHostHandler).Methods("GET")
router.HandleFunc("/hosts", CreateHostHandler).Methods("POST")
router.HandleFunc("/hosts/{id}", UpdateHostHandler).Methods("PATCH", "PUT")
router.HandleFunc("/hosts/{id}", DeleteHostHandler).Methods("DELETE")
```

For singleton resources:
```
router.HandleFunc("/install-config", GetInstallConfig).Methods("GET")
router.HandleFunc("/install-config", CreateInstallConfig).Methods("POST")
router.HandleFunc("/install-config", UpdateInstallConfig).Methods("PATCH", "PUT")
router.HandleFunc("/install-config", DeleteInstallConfig).Methods("DELETE")
```

Handlers for each resource could be held in a separate module. We could even omit `...Handler` in the handler function name.

honza:
@hardys I wonder if this is no longer relevant now that kni-installer is in play?

knowncitizen:
> @hardys I wonder if this is no longer relevant now that kni-installer is in play?

Adding @markmc for comment as well

markmc:
Currently in dev-scripts we do:

```
cat >> ocp/install-config.yaml <<EOF
...
EOF
kni-install --dir ocp create ignition-configs
cp ocp/master.ign ocp/master.ign.tmp
kni-install --dir ocp create cluster
cp ocp/master.ign.tmp ocp/master.ign

# lots of stuff here to provision the master nodes using master.ign

# wait for the master API to become available

oc --config ocp/auth/kubeconfig get nodes

# We're done!
```

and then use the `ocp/master.ign` to provision the master nodes. So facet should do the same as this ... for now.

We will soon be working on moving the master node provisioning into `kni-install` itself, and then the above becomes simply:

```
cat >> ocp/install-config.yaml <<EOF
...
EOF
kni-install --dir ocp create cluster

# wait for the master API to become available

oc --config ocp/auth/kubeconfig get nodes

# We're done!
```

So, no - I don't think a separate API for the "create ignition-configs" makes sense long-term

Thanks for asking!

honza:
Alright, closing this as stale.

PR #27, opened 2019-02-14
Connect hosts to api

jtomasek:
Populates HostsTable with actual data from API.

Work in progress
- needs to add reselect for cached selectors
- needs to add loading status tracking
- clean up HostsExample from App.tsx

Depends on #22


src/models/hosts.ts
```diff
@@ -1,3 +1,9 @@
 export interface Host {
-  id: string;
+  name: string;
+  ip: string;
```
honza: The types of the fields of `Host` should match whatever the server returns.

src/components/BaremetalInventory.tsx
```diff
@@ -1,37 +1,66 @@
...
-      variant={PageSectionVariants.light}
-      isMain
-      style={{ padding: 0 }}
-    >
-      <HostsTable />
-    </PageSection>
-  </Fragment>
-);
+interface Props {
+  hostRows: string[][];
```
honza: Let's use a more concrete type here instead of `string[][]`

src/components/HostsTable.tsx
```diff
@@ -1,8 +1,22 @@
 import React, { FC, Fragment } from 'react';
 import { Table, TableHeader, TableBody } from '@patternfly/react-table';
+import { Bullseye } from '@patternfly/react-core';
 
-const HostsTable: FC = (): JSX.Element => {
-  const headerStyle = { position: 'sticky', top: 0, background: 'white' };
+interface Props {
+  hostRows: string[][];
```
honza: Same as above

jtomasek:
Rebased on top of `master`

src/components/HostsTable.tsx
jtomasek: done

src/components/BaremetalInventory.tsx
jtomasek: done

src/models/hosts.ts
jtomasek: done

PR #31, opened 2019-02-18
Implement cluster wizard bottom navigation

jtomasek:
* Simplify App component by extracting PageSidebar
* Add redux-devtools-extension
* Add clusterWizard state management
* Extract ClusterWizardNav into reusable component and move
      the specific contents to respective wizard steps components
* Introduce ToolbarProgress, ToolbarButton and ToolbarText components

Depends on #27


lizsurette:
@jtomasek nice to see things clickable! I noticed that the buttons at the bottom are in reverse order than what PatternFly calls for. Would you mind swapping "Deploy Cluster" with "Back" so that the primary button is left aligned?

src/components/ui/Toolbar/ToolbarProgress.tsx
```diff
@@ -0,0 +1,22 @@
+import React, { FC } from 'react';
+
+import {
+  ToolbarGroup,
+  ToolbarItem,
+  Progress,
+  ProgressSize
+} from '@patternfly/react-core';
+
+const ToolbarProgress: FC<React.ComponentProps<typeof Progress>> = ({
```
honza: This is cool. I like this component.

src/reducers/hosts.ts
```diff
@@ -15,7 +15,7 @@ export default combineReducers<HostsState, HostsActions>({
   hosts: (state = [], action) => {
     switch (action.type) {
       case getType(hosts.fetchHosts.success):
-        return [...state, ...action.payload];
+        return [...action.payload];
```
honza: Good catch.

PR #33, opened 2019-02-18
Added information about using Prettier to the README

knowncitizen:


README.md
```diff
@@ -141,6 +141,10 @@ In the project directory, you can run:
 
 Installs dependencies to node_modules directory
 
+### `yarn prettier`
+
+This application uses Prettier to check and format code. You can run the above command to clean your code, or if you're using something like VSCode, you can set up the Prettier extenson and formatting changes will automatically be applied when you save.
```
honza: Where is the love for Emacs? :smile: I would prefer a [link to the documentation](https://prettier.io/docs/en/editors.html) on how to integrate it into your own editor.

Also, please wrap the line to something more manageable. 252 chars is a bit unwieldy.

knowncitizen: I disagree with the line wrap - for something like a README markdown, it properly wraps when viewed in something like Github.

honza: The whole README is a joke. :(

knowncitizen: We should probably split some of it off and fix that which has been line-wrapped.

honza: I think markdown files should be hardwrapped at 79 characters. GitHub converts markdown files to HTML, removing any hard newlines in the process. The original README was created via CRA, and for some reason they don't respect that common practice. I feel strongly (though not violently) that this is the right way (tm).

However, if you disagree, I'm not going to block your PRs. This was previously indicated by my approval of this present PR.

knowncitizen: I think hard wrapping at 79 characters is a bit of a deprecated ideology, tbh. That was designed for when we had 80 column CRTs :) Other than that fact, I'm pretty agnostic.  @jtomasek @flofuchs @belwell @dantrainor  thoughts? Let's decide here and close it out.

honza: I don't think it is. Even novels in printed books have columns of width 2-3 English alphabets (rule of thumb). Reading long lines of text is harder because the eyes has to track back to the beginning of the line to read the next line. If the line is shorter, this is easier and less tiring.

Also, we might have wider and bigger monitors now, but that doesn't mean we should write files that are hundreds of characters in width. Instead, if your files are wrapped at around 79 chars, you can fit multiple files side by side on your large monitor (3 in my case). I don't have to keep scrolling to the end of the line.

knowncitizen: Entirely fair. I'm 100% open to wrapping at 79 characters.

knowncitizen: Can we configure Prettier for markdown to yell/wrap at 79?

honza: Two steps:

* Add the README to the list of files that prettier checks
* Pass `--prose-wrap always` flag to the command

knowncitizen: @honza I'm not sure what to do here. I did both of those things - in `package.json`, I added:     ```"prettier": "prettier --write './*.{md}' --write 'src/**/*.{js,jsx,ts,tsx,md}'",```
And I added the proseWrap to the rc. When I run `npm run prettier` the md file doesn't show up in the list. What am I missing?

honza: The `--write` flag tells prettier to write the file in place after applying the transformations. I don't think you need the `--write './*.{md}` part. The path at the end looks for md files in the src directory. I'm not sure off the top of my head how to specify the path correctly here.

knowncitizen: I should have mentioned that I tried it several ways, never getting the md file to be recognized. Very odd... will try more today.

honza: Don't worry about it if it's too complicated or taking too long. It's kind of low priority. :wink:

knowncitizen: I'm not putting too much time on it :) Just want it to work...

knowncitizen:
Rebased, added config info here: https://github.com/metalkube/facet/pull/38

honza:
I don't think the rebase is quite right.

PR #34, opened 2019-02-19
Reimplement GetHosts with k8s types

honza:
The actual types are coming from metalkube/baremetal-operator.  These
datastructures represent the Kubernetes resources for Day 2.  We are
adding them here for reusability, and compatibility between Day 1 and
Day 2.


honza:
Here is an example response:

```json
{
  "data": {
    "metadata": {
      "selfLink": "/api/v1/namespace/default/baremetalhosts",
      "resourceVersion": "123456"
    },
    "items": [
      {
        "metadata": {
          "name": "baremetalhost",
          "namespace": "default",
          "uid": "some-uuid",
          "creationTimestamp": "2019-01-22T09:30:00Z",
          "labels": {
            "app": "hello-world"
          }
        },
        "spec": {
          "bmc": {
            "ip": "192.168.100.100",
            "credentialsName": "bmc-creds-valid"
          },
          "online": true
        },
        "status": {
          "lastUpdated": "2019-02-19T19:32:26Z",
          "hardware": {
            "ramGiB": 128,
            "nics": [
              {
                "mac": "00:A0:C9:14:C8:29",
                "ip": "192.168.100.100",
                "speedGbps": 40
              }
            ],
            "storage": [
              {
                "sizeGiB": 1024,
                "info": "disk info"
              }
            ],
            "cpus": [
              {
                "type": "intel",
                "speedGHz": 3
              }
            ]
          },
          "provisioningID": "some ironic ID",
          "image": "image.qcow",
          "goodCredentials": null,
          "errorMessage": ""
        }
      }
    ]
  }
}
```

dhellmann:
We should probably expose a method on the host to return its status, instead of expecting users of the class to look at the label directly.

dhellmann:
> We should probably expose a method on the host to return its status, instead of expecting users of the class to look at the label directly.

See https://github.com/metalkube/baremetal-operator/pull/15

honza:
@dhellmann With the Gopkg.lock change in the operator repo, this PR now imports that code correctly. :100:

pkg/server/handler_helpers.go
```diff
@@ -33,6 +33,15 @@ func RespondWithJson(w http.ResponseWriter, obj interface{}) {
 	}
 }
 
+func RespondWithK8s(w http.ResponseWriter, obj interface{}) {
+	err := json.NewEncoder(w).Encode(obj)
```
jtomasek: Can we do this in default RespondWithJson? nesting under `data` seems not necessary.

honza: Yes, I agree.

My original thought was that each response would have a "status" and a "data" to make it uniform.

honza:
@jtomasek The build is now failing because of two things:

1. prettier and eslint seem to be in disagreement about the indentation
2. our eslint config isn't compatible with that of openshift/console and won't allow the use of `any`

Do you have any thoughts/ideas on how to do this? Should we relax our rules to be more inline with them? Or should we reimplement the types ourselves more properly/strictly?

honza:
I disabled the troublesome rules in that file only. We should try and keep openshift/console code in separate files so we can ignore these types of errors in isolation.

jmolmo:
Maybe a silly question because probably i'm missing something ... 
Wouldn't it be good to have fields for showing status/errorMessage of each of the different hardware devices? for example, If we have errors from several devices, i only see one place to put this information (status.errorMessage). Maybe to mix all this information here will result confuse for the final user..

honza:
@jmolmo This might be better directed towards the baremetal-operator project where these values are defined. We're simply using it as a dependency.

dhellmann:
@jmolmo it's all still early days. The current data structure is just a placeholder until we know what we really care about. The baremetal operator isn't going to be monitoring devices for errors, though, so we may end up with some other resources to hold more detailed status if we need a place to put errors for individual drives or whatever.

src/selectors/hosts.ts
```diff
@@ -7,7 +7,21 @@ export const getHosts = (state: RootState): Host[] => state.hosts.hosts;
 export const getHostsLoading = (state: RootState): boolean =>
   state.hosts.loading;
 
+const hostToHostTableRow = (host: Host): string[] => {
+  const spec = host.spec ? host.spec : {};
+  const status = host.status ? host.status : {};
```
jtomasek: these 2 lines can be shortened to:   `const { spec = {}, status = {} } = host;`

pkg/server/handler_helpers.go
jtomasek: Do you want to update RespondWithJson to match this in follow up patch?

src/selectors/hosts.ts
honza: Done

pkg/integration/hosts.go
```diff
@@ -15,7 +15,10 @@
 package integration
 
 import (
+	apis "github.com/metalkube/baremetal-operator/pkg/apis/metalkube/v1alpha1"
```
knowncitizen: is this arbitrary?

honza: I'm not sure what you mean.

knowncitizen: Just wondering if it follows a convention.

honza: It most likely does. But this is a link to the actual package. We're only consuming it.

PR #35, opened 2019-02-20
Add "host" CLI command

honza:
See `README.md` for more details

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,13 +75,16 @@ You can use the binary directly:
 
 ```
 ./bin/facet -h
-facet server
+Facet
+
+Metalkube Facet is an interface to kubernetes baremetal provisioning.
 
 Usage:
   facet [command]
 
 Available Commands:
   help        Help about any command
+  host        Interact with baremetal hosts
   server      Run the facet server
 
 Flags:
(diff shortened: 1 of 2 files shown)


hardys:
Thanks this is a nice example, I pulled/tested it and it works well.

We should be able to add the gophercloud calls behind this pretty easily so we can return real data from ironic, although it'd be nice if we could do that in a way that would be easily reusable for the baremetal operator.

One comment/question was around the CLI command groups - where possible I think we'll want to align with openshift-install, e.g for ignition-configs it's `openshift-install create ignition-configs`, should we namespace hosts so we do e.g `facet create hosts` or something?

That's just a nit at this point, I'd be happy to see this land as-is, but I think in due course it would be nice if the interface became a superset of openshift-install, sharing code where practical and making it easier if/when we want to integrate baremetal management into openshift-install directly?

hardys:
One other name-bikeshed nit, I'd imagined we'd call this metalkube-install or something, but facet is fine - I just don't find it a particularly self-documenting name even though it's a clever play on words :)

honza:
> I just don't find it a particularly self-documenting name even though it's a clever play on words :)

You'll have to talk to @knowncitizen about that one. :smile: 

> should we namespace hosts so we do e.g `facet create hosts` or something?

I'm happy to work on that in subsequent patches. I don't really have a strong opinion one way or another.

honza:
* Rebase on master

knowncitizen:
@hardys I don't have a particularly strong opinion of 'facet' vs. 'metalkube-install'.  If you or anyone else wants to change it, I'm not going to fight it. I can definitely see your side since it's not self-documenting.

PR #36, opened 2019-02-21
Introduce pull secret fetching

jtomasek:
* Converts cluster definition form to using [Formik](https://jaredpalmer.com/formik/)
* Adds basic fields validation
* Introduces capabilities to submit the form to the respective API endpoint
* Introduces /api/install-config API endpoint

TODO:
* [x] Move install config handler(s) into a separate module
* [x] Move pullSecret and developers.redhat.com token fetching into a separate functions
* [x] Display error in the footer

Separate PR:
* Add client side validation for clusterName, DNSDomain, and pullSecret format
* Add state management for InstallConfig
* Persist install config on server side
* Add GET /api/install-config endpoint and handler to allow retrieving stored install config

### DIFF
--- a/pkg/server/handlers.go
+++ b/pkg/server/handlers.go
@@ -15,11 +15,21 @@
 package server
 
 import (
-	"github.com/metalkube/facet/pkg/common"
-	"github.com/metalkube/facet/pkg/integration"
+	"encoding/json"
 	"log"
 	"net/http"
+
+	"github.com/metalkube/facet/pkg/common"
+	"github.com/metalkube/facet/pkg/integration"
 )
+
+type ClusterDefinition struct {
+	ClusterName string `json:"cluster_name"`
+	DNSDomain   string `json:"dns_domain"`
+	Username    string `json:"username"`
+	Password    string `json:"password"`
+	PullSecret  string `json:"pull_secret"`
+}
 
 func HostsHandler(w http.ResponseWriter, r *http.Request) {
 	hostList, err := integration.GetHosts()
@@ -28,6 +38,36 @@ func HostsHandler(w http.ResponseWriter, r *http.Request) {
 		log.Print(err)
 	}
 	RespondWithJson(w, hostList)
+}
+
+func CreateClusterDefinition(w http.ResponseWriter, r *http.Request) {
+	var clusterDefinition ClusterDefinition
+	err := json.NewDecoder(r.Body).Decode(&clusterDefinition)
+	if err != nil {
+		RespondWithError(w, err, 500)
+		return
+	}
+
+	if clusterDefinition.PullSecret != "" {
+		// TODO(jtomasek): update install-config.yaml with clusterDefinitionData (except username/password)
+		RespondWithJson(w, clusterDefinition.PullSecret)
+	} else {
+		tokenData, err := integration.FetchAuthToken(clusterDefinition.Username, clusterDefinition.Password)
+		if err != nil {
+			RespondWithError(w, err, 500)
+			return
+		}
+
+		pullSecret, err := integration.FetchPullSecret(tokenData)
+		if err != nil {
+			RespondWithError(w, err, 500)
+			return
+		}
+
+		// TODO(jtomasek): update install-config.yaml with clusterDefinitionData (except username/password)
+		// TODO(jtomasek): Response should be a complete InstallConfig object
+		RespondWithJson(w, pullSecret)
+	}
 }
 
 // This is an example of a REST API endpoint handler which will trigger a long
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -15,10 +15,11 @@
 package server
 
 import (
-	"github.com/metalkube/facet/pkg/common"
-	_ "github.com/metalkube/facet/statik"
 	"log"
 	"net/http"
+
+	"github.com/metalkube/facet/pkg/common"
+	_ "github.com/metalkube/facet/statik"
 )
 
 type Server struct {
@@ -53,8 +54,7 @@ func (s *Server) Start() {
 	err := http.ListenAndServe(":"+s.Port, router)
 
 	if err != nil {
-		log.Fatal("Failed to start server")
-		log.Fatal(err)
+		log.Fatal("Failed to start server: ", err)
 	}
 
 }
--- a/src/components/createCluster/validationSchema.ts
+++ b/src/components/createCluster/validationSchema.ts
@@ -1,0 +1,18 @@
+import * as Yup from 'yup';
+
+export default Yup.object().shape({
+  clusterName: Yup.string().required('Required'),
+  DNSDomain: Yup.string().required('Required'),
+  pullSecret: Yup.string().when('$providePullSecret', {
+    is: true,
+    then: Yup.string().required('Required')
+  }),
+  username: Yup.string().when('$providePullSecret', {
+    is: false,
+    then: Yup.string().required('Required')
+  }),
+  password: Yup.string().when('$providePullSecret', {
+    is: false,
+    then: Yup.string().required('Required')
+  })
+});
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@
 
 # misc
 .DS_Store
+.vscode
 .env.local
 .env.development.local
 .env.test.local
--- a/package.json
+++ b/package.json
@@ -8,14 +8,18 @@
     "@patternfly/react-icons": "^3.0.0",
     "@patternfly/react-table": "0.4.18",
     "@types/jest": "23.3.13",
+    "@types/lodash": "^4.14.121",
     "@types/node": "10.12.18",
     "@types/react": "16.7.20",
     "@types/react-dom": "16.0.11",
     "@types/react-redux": "^7.0.1",
+    "@types/yup": "^0.26.9",
     "axios": "^0.18.0",
     "bootstrap-sass": "^3.4.0",
     "eslint": "5.6.0",
     "font-awesome": "^4.7.0",
+    "formik": "^1.5.1",
+    "lodash": "^4.17.11",
     "node-sass": "^4.11.0",
     "patternfly": "^3.59.1",
     "patternfly-react": "^2.29.10",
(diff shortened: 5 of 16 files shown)


pkg/server/handlers.go
```diff
@@ -30,6 +35,64 @@ func HostsHandler(w http.ResponseWriter, r *http.Request) {
...
+	DNSDomain   string `json:"DNSDomain"`
+	PullSecret  string `json:"pullSecret"`
+}
+
+type DefineClusterCredentials struct {
+	ClusterName string `json:"clusterName"`
+	DNSDomain   string `json:"DNSDomain"`
+	Username    string `json:"username"`
+	Password    string `json:"password"`
+}
```
hardys: This is a bit confusing, as it conflates the ClusterName/Domain with the credentials required to get the pull secret (which aren't the credentials needed to access the cluster?)

pkg/server/handlers.go
```diff
@@ -30,6 +35,64 @@ func HostsHandler(w http.ResponseWriter, r *http.Request) {
...
+
+func CreateInstallConfig(w http.ResponseWriter, r *http.Request) {
+	var clusterDefinition DefineClusterCredentials
+	_ = json.NewDecoder(r.Body).Decode(&clusterDefinition)
```
honza: Ignored error:

```go
err := json.NewDecoder(r.Body).Decode(&clusterDefinition)
if err := nil {
    RespondWithError(w, err, 400)
    return
}
```

pkg/server/handlers.go
```diff
@@ -30,6 +35,64 @@ func HostsHandler(w http.ResponseWriter, r *http.Request) {
...
+
+	tokenRequestURL := fmt.Sprintf("https://developers.redhat.com/auth/realms/rhd/protocol/openid-connect/token")
```
honza: `fmt.Sprintf` seems unnecessary here --- is this a left-over?

Also, the URL could be a constant at the top of the file?

pkg/server/handlers.go
```diff
@@ -30,6 +35,64 @@ func HostsHandler(w http.ResponseWriter, r *http.Request) {
...
+
+	resp, err := http.PostForm(tokenRequestURL, url.Values{
+		"grant_type": {"password"},
+		"client_id":  {"uhc"},
+		"username":   {clusterDefinition.Username},
+		"password":   {clusterDefinition.Password},
+	})
+
+	if err != nil {
+		log.Print("Error in api call", err)
```
honza: After logging the error, we should return an error to the client. Also, later code will fail as the `resp` variable is unpopulated. 

```go
RespondWithError(w, err, 500)
return
```

pkg/server/handlers.go
```diff
@@ -30,6 +35,64 @@ func HostsHandler(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	// Is this needed?
+	// defer resp.Body.Close()
+
+	var tokenData map[string]string
+	json.NewDecoder(resp.Body).Decode(&tokenData)
+
+	if err != nil {
+		RespondWithError(w, err, 500)
```
honza: And return right after the response.

pkg/server/handlers.go
```diff
@@ -30,6 +35,64 @@ func HostsHandler(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	pullSecretRequestURL := fmt.Sprintf("https://api.openshift.com/api/accounts_mgmt/v1/access_token")
```
honza: Same as above

pkg/server/handlers.go
```diff
@@ -30,6 +35,64 @@ func HostsHandler(w http.ResponseWriter, r *http.Request) {
...
+
+	req, err := http.NewRequest("POST", pullSecretRequestURL, bytes.NewReader(nil))
```
honza: Moar error handling. This is one of the most hated features of Go: the constant `if err != nil {}`. But it actually forces you to think about your errors.

pkg/server/handlers.go
```diff
@@ -30,6 +35,64 @@ func HostsHandler(w http.ResponseWriter, r *http.Request) {
...
+	req.Header.Add("accept", "application/json")
+	req.Header.Add("authorization", "bearer "+tokenData["access_token"])
+	resp, err = http.DefaultClient.Do(req)
+
+	// Is this needed?
+	// defer resp.Body.Close()
+
+	var pullSecret map[string]interface{}
+	json.NewDecoder(resp.Body).Decode(&pullSecret)
```
honza: This could fail.

pkg/server/server.go
```diff
@@ -53,8 +54,8 @@ func (s *Server) Start() {
 	err := http.ListenAndServe(":"+s.Port, router)
 
 	if err != nil {
-		log.Fatal("Failed to start server")
 		log.Fatal(err)
```
honza: The process will actually die once it encounters a `log.Fatal` call.

jtomasek:
Updated the error handling, moved CreateClusterDefinition handler into a separate file.

pkg/server/cluster_definition_handler.go
```diff
@@ -0,0 +1,95 @@
...
+	ExpiresIn        int    `json:"expires_in"`
+	RefreshExpiresIn int    `json:"refresh_expires_in"`
+	RefreshToken     string `json:"refresh_token"`
+	TokenType        string `json:"token_type"`
+	NotBeforePolicy  int    `json:"not-before-policy"`
+	SessionState     string `json:"session_state"`
+	AccessToken      string `json:"access_token"`
+}
+
+func fetchAuthToken(Username string, Password string) (tokenData *DeveloperTokenResponse, err error) {
```
honza: Lowercase names for function arguments

pkg/server/cluster_definition_handler.go
```diff
@@ -0,0 +1,95 @@
...
+		"grant_type": {"password"},
+		"client_id":  {"uhc"},
+		"username":   {Username},
+		"password":   {Password},
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	err = json.NewDecoder(resp.Body).Decode(&tokenData)
```
honza: Before reading the data, you need to check the status code. The above error check only tells you if the HTTP request succeeded. It may have been successful, but the response could have been 401.

```go
if resp.StatusCode != 200 {
    return nil, errors.New("Unauthorized")
}
```

pkg/server/cluster_definition_handler.go
honza: Return a value, not a pointer

pkg/server/cluster_definition_handler.go
```diff
@@ -0,0 +1,95 @@
...
+
+	err = json.NewDecoder(resp.Body).Decode(&pullSecret)
+	if err != nil {
+		return nil, err
+	}
+
+	return pullSecret, nil
+}
+
+func CreateClusterDefinition(w http.ResponseWriter, r *http.Request) {
```
honza: I think the handler should be in `handlers.go`, and the other code should be in `pkg/integration/token` (or similar). That way the CLI can also get the pull secret.

src/components/createCluster/validationSchema.ts
```diff
@@ -0,0 +1,20 @@
...
+  DNSDomain: Yup.string().required('Required'),
+  pullSecret: Yup.string().when('$providePullSecret', {
+    is: true,
+    then: Yup.string().required('Required')
+  }),
+  username: Yup.string().when('$providePullSecret', {
+    is: false,
+    then: Yup.string()
+      .required('Required')
+      .email('Incorrect format, please provide email')
```
honza: My username isn't an email. I wasn't able to submit the form without tweaking this.

pkg/server/cluster_definition_handler.go
```diff
@@ -0,0 +1,95 @@
...
+	Password    string `json:"password"`
+	PullSecret  string `json:"pullSecret"`
+}
+
+type DeveloperTokenResponse struct {
+	ExpiresIn        int    `json:"expires_in"`
+	RefreshExpiresIn int    `json:"refresh_expires_in"`
+	RefreshToken     string `json:"refresh_token"`
+	TokenType        string `json:"token_type"`
+	NotBeforePolicy  int    `json:"not-before-policy"`
```
honza: This should use underscores for consistency.

pkg/server/cluster_definition_handler.go
```diff
@@ -0,0 +1,95 @@
...
+	"encoding/json"
+	"net/http"
+	"net/url"
+)
+
+const authTokenRequestURL = "https://developers.redhat.com/auth/realms/rhd/protocol/openid-connect/token"
+const pullSecretRequestURL = "https://api.openshift.com/api/accounts_mgmt/v1/access_token"
+
+type ClusterDefinition struct {
+	ClusterName string `json:"clusterName"`
```
honza: Snake case might be better than camel case.

pkg/server/cluster_definition_handler.go
jtomasek: Done

pkg/server/cluster_definition_handler.go
jtomasek: Done, it will need to get enhanced so the error returned includes status code etc. rather than just a string.

pkg/server/cluster_definition_handler.go
jtomasek: Done

src/components/createCluster/validationSchema.ts
jtomasek: Removed email validation for now

pkg/server/cluster_definition_handler.go
jtomasek: This format comes from developer.redhat.com, I am not sure I can do anything about it.

pkg/server/cluster_definition_handler.go
jtomasek: Done, although it requires pre-processing the data on the JS side. Unless there is a possibility to set json flags same as we do in Go, which I don't know about.

pkg/server/cluster_definition_handler.go
honza: :cry:

jtomasek:
Tests are failing due to this issue: https://github.com/facebook/create-react-app/issues/5241. I'll replace lodash-es with lodash and use named exports for now.

PR #38, opened 2019-02-26
Updated prettier config files

knowncitizen:
- rewrote prettier initiation in package.json
 - added .prettierignore for vendor files (and anything else)
 - added proseWrap: always to prettierrc
 - committed changes from a prettier run


package.json
```diff
@@ -36,7 +36,7 @@
     "build": "react-scripts build",
     "test": "react-scripts test",
     "eject": "react-scripts eject",
-    "prettier": "prettier --write 'src/**/*.{js,jsx,ts,tsx}'",
+    "prettier": "prettier --write \"**/*.{js,jsx,ts,tsx,json,md,scss,yaml,yml}\"",
```
jtomasek: By keeping the single quotes you could avoid having to escape them.

.prettierignore
```diff
@@ -0,0 +1 @@
+vendor/
```
jtomasek: `build/` should be also ignored IMHO

knowncitizen: Done

package.json
knowncitizen: Done

PR #39, opened 2019-02-27
WIP Create selector to agregate values from all hosts

belwell:


src/selectors/hosts.ts
```diff
@@ -11,3 +11,19 @@ export const getHostTableRows = createSelector(
   getHosts,
   hosts => hosts.map(host => Object.values(host))
 );
+
+const hostsValuesSelector = state => state.hosts.items;
```
jtomasek: Please check the changes which PR#34 introduces. It defines the actual structure of the host. So you can match the selectors to it to save time reimplementing the selectors later.
https://github.com/metalkube/facet/pull/34/files#diff-d75362b21270945abccac41ae9de67c5

jtomasek: Your selector should be based on hosts data from `getHosts`.  hostsValuesSelector should not be needed.

src/selectors/hosts.ts
```diff
@@ -11,3 +11,19 @@ export const getHostTableRows = createSelector(
...
+
+export const hostsAggregateValues = createSelector(
+  hostsValuesSelector,
+  items => items.reduce((acc, item) => acc + item.memory, 0)
```
jtomasek: This looks good, I assume that `acc` is going to be an object of an explicit type which holds the aggregate values. For each host, aggregate values will be increased by the values from the host.

PR #41, opened 2019-03-04
Open up backend server to hosts other than localhost

dantrainor:
While the Yarn development server listens on localhost and the
primary system interface, the Facet backend server does not.
Open up the backend server to listen on all interfaces in a
manner similar to the Yarn development server, to aid in
remote development.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,5 +61,5 @@
     "eslint-plugin-react": "^7.12.4",
     "prettier": "1.16.1"
   },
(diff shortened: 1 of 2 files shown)


honza:
This doesn't change the interface on which the backend server is listening on. You have to change the interface in the server package.

honza:
> Could you please rebase your commits with master branch and avoid master branch merge commits

If the content of the PR is fine, then we can just approve and squash.

PR #43, opened 2019-03-07
Facet Architecture diagram

knowncitizen:


hardys:
Looks good, but I think we should replace the reference to libvirt with kni-install, since that's the abstraction we'll use to launch the VM from the API, not libvirt-go directly?

honza:
Yes, the diagram predates the kni-installer discussions.

knowncitizen:
Sounds good. I'll change the diagram.

PR #50, opened 2019-03-13
Fixed build after move of facet under openshift-metalkube org

booxter:


booxter:
Nevermind, it's a dup of https://github.com/openshift-metalkube/facet/pull/49

PR #53, opened 2019-03-25
Proper 'listen' message

e-minguez:


e-minguez:
> The `ListenAndServe` call blocks while it's running. This patch would hide the message until the server process is terminated.
> 
> If the server cannot be started. the call in question will return nearly immediately with an error message.

Done

PR #56, opened 2019-04-24
fix metalkube to metal3 rename

dhellmann:


knowncitizen:
I don't want to change all of the readmes back, I pulled in just your hosts.go changes in #57

PR #64, opened 2020-04-27
New host actions: enable/disable

mareklibra:


mareklibra:
01_disable
---
02_disableError
---
04_disabled
---
05_enabled
---
06_enabling

src/components/clusterWizard/HostsTable.scss
```diff
@@ -0,0 +1,3 @@
+.host-table-alerts {
+  padding-top: 0.5rem !important;
```
jtomasek, line 2: I think we should use variable
```suggestion
  padding-top: var(--pf-global--spacer--lg) !important;
```

Also the file does not have to be `scss`, we most likely won't need any scss features and can use plain css.

src/components/clusterWizard/HostsTable.tsx
```diff
@@ -116,17 +131,90 @@ const HostsTable: React.FC<HostsTableProps> = ({ hosts = [], uiState, fetchHosts
     [hostRows, openRows],
   );
 
+  const addAlert = React.useCallback(
+    (alert: Alert) => {
+      if (alert.key) {
```
jtomasek, line 136: When adding an alert with the same key which already exists, do we want to replace the alert?

src/components/clusterWizard/HostsTable.scss
mareklibra: Renamed

src/components/clusterWizard/HostsTable.tsx
mareklibra: Good point. With recent implementation, the alert is not duplicated but moved to the top instead.

PR #66, opened 2020-04-28
Cluster wizard form

jtomasek:
Screenshot 2020-04-28 at 07 51 37


src/components/clusterWizard/ClusterWizardForm.tsx
```diff
@@ -8,67 +9,73 @@ import {
...
+    serviceNetworkCIDR: cluster.serviceNetworkCIDR || '',
+    apiVip: cluster.apiVip || '',
+    dnsVip: cluster.dnsVip || '',
+    ingressVip: cluster.ingressVip || '',
+    pullSecret: cluster.pullSecret || '',
+    sshPublicKey: cluster.sshPublicKey || '',
+    hostsRoles:
+      cluster.hosts?.map((host: Host) => ({
+        id: host.id,
+        role: host.role as 'master' | 'worker',
```
mareklibra, line 51: Worth of it's own type. List of these roles is expected to be changed in the future.

src/components/clusterWizard/ClusterWizardForm.tsx
```diff
@@ -8,67 +9,73 @@ import {
...
-        formikActions.setStatus({ error: e.message });
-        formikActions.setSubmitting(false);
-        // TODO(jtomasek): dispatch a failure action
-      });
+    try {
+      const { data } = await patchCluster(cluster.id, values); // eslint-disable-line @typescript-eslint/no-non-null-assertion
+      console.log('data', data);
+      // TODO(jtomasek): update cluster in redux
+    } catch (e) {
+      formikActions.setStatus({ error: 'Failed to update the cluster.' });
```
mareklibra, line 75: Can we reuse `handleApiError()` from #65?

src/components/clusterWizard/ClusterWizardForm.tsx
```diff
@@ -89,63 +103,131 @@ const ClusterWizardForm: React.FC<ClusterWizardFormProps> = ({ cluster, setStep
                     <TextContent>
                       <Text component="h1">Configure a bare metal OpenShift cluster</Text>
                     </TextContent>
+                    <InputField label="Cluster name" name="name" isRequired />
+                    <SelectField
+                      label="OpenShift Version"
+                      name="openshiftVersion"
+                      options={[
+                        { label: 'OpenShift 4.4', value: '4.4' },
```
mareklibra, line 111: Can we externalize these constants? And use the first of them to feed the default of initalValue in `ClusterWizardForm.tsx`?

src/components/clusterWizard/ClusterWizardForm.tsx
```diff
@@ -89,63 +103,131 @@ const ClusterWizardForm: React.FC<ClusterWizardFormProps> = ({ cluster, setStep
...
-                    <TextAreaField name="sshPublicKey" label="SSH Public Key" isRequired />
+                    <TextAreaField
+                      name="pullSecret"
+                      label="Pull Secret"
+                      helperText={
+                        <>
+                          The pull secret obtained from the Pull Secret page on the{' '}
+                          {
+                            <a
+                              href="https://cloud.redhat.com/openshift/install/metal/user-provisioned"
```
mareklibra, line 174: A constant again. I think we should introduce a file with just these "config-like" constants.

src/components/clusterWizard/ClusterWizardForm.tsx
```diff
@@ -89,63 +103,131 @@ const ClusterWizardForm: React.FC<ClusterWizardFormProps> = ({ cluster, setStep
...
+                title={status.error}
+                action={<AlertActionCloseButton onClose={() => setStatus({ error: null })} />}
+              />
+            )}
+          </AlertsSection>
           <ClusterWizardToolbar>
             <ToolbarButton
               variant={ButtonVariant.secondary}
-              onClick={() => setStep(WizardStep.BaremetalInventory)}
+              component={(props) => <Link to="/clusters" {...props} />}
```
mareklibra, line 209: What about referring the URL via a constant? We are still discussing the future deployment of the app and this might be affected.

src/components/clusterWizard/ClusterWizardForm.tsx
jtomasek: I'll remove the hostRoles definition here for now as it is not needed.

src/components/clusterWizard/ClusterWizardForm.tsx
jtomasek: The transition in the wizard is currently done via state not router, the `/cluster` route points to `ClusterPage` component.

At some point we can decide to drive it via routes but I kept it in state for now so we're in control of how we allow the user to transition between the steps.

jtomasek: I agree we should turn the routes into constants (`config/routes` maybe?) probably in separate PR?

PR #67, opened 2020-04-28
Enrich HostsTable for selection of a role

mareklibra:
The user can newly select role of the host within HostsTable.

TODO:
- either force UI refresh or update redux by updated cluster (after patch) - as a follow-up

### DIFF
--- a/src/components/clusterWizard/RoleDropdown.tsx
+++ b/src/components/clusterWizard/RoleDropdown.tsx
@@ -1,0 +1,38 @@
+import React from 'react';
+import { Host, ClusterUpdateParams } from '../../api/types';
+import { SimpleDropdown } from '../ui/SimpleDropdown';
+import { patchCluster } from '../../api/clusters';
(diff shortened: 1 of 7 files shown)


src/components/clusterWizard/RoleDropdown.tsx
```diff
@@ -0,0 +1,38 @@
...
+    const params: ClusterUpdateParams = {};
+    setDisabled(true);
+    params.hostsRoles = [
+      {
+        id: host.id,
+        role: role as HostRolesType,
+      },
+    ];
+    // TODO(mlibra): store updatedCluster (infra WIP) or initiate reload
+    // const updatedCluster =
```
mareklibra, line 25: Will be stored to redux and hosts will be retrieved from Cluster (not as recently via specific request). WIll be Follow-up for https://github.com/openshift-metal3/facet/pull/71

mareklibra:
Rebased

PR #68, opened 2020-04-29
Use redux toolkit to handle clusters actions

jtomasek:
- add @reduxjs/toolkit dependency
- remove unused types, reducers, actions, selectors
- reimplement clusters component to use `clusterSlice`


src/components/clusters/Clusters.tsx
```diff
@@ -38,15 +34,14 @@ const Clusters: React.FC<ClustersProps> = ({
...
-  clustersError,
 }) => {
   React.useEffect(() => {
     fetchClusters();
   }, [fetchClusters]);
 
   const errorState = (
     <PageSection variant={PageSectionVariants.light} isMain>
-      <ErrorState title={clustersError} fetchData={fetchClusters} />;
+      <ErrorState title="Failed to fetch clusters." fetchData={fetchClusters} />;
```
mareklibra, line 44: Can we render text passed to the promise's reject function? It will be more flexible this way.

mareklibra:
I like the `@reduxjs/toolkit`

PR #71, opened 2020-04-29
Add polling to ClusterPage

mareklibra:
The single-cluster page polls for data update.

Call
```
  dispatch(forceReloadAction(true))
```
to force refresh immediately.

### DIFF
--- a/src/components/clusters/ClusterPage.tsx
+++ b/src/components/clusters/ClusterPage.tsx
@@ -18,30 +18,41 @@ type MatchParams = {
   clusterId: string;
 };
 
-const ClusterPage: React.FC<RouteComponentProps<MatchParams>> = ({ match }) => {
-  const { clusterId } = match.params;
-  const { data: cluster, uiState, forceReload } = useSelector(selectCurrentClusterState);
+const useFetchCluster = (clusterId: string) => {
   const dispatch = useDispatch();
+  return React.useCallback(() => dispatch(fetchClusterAsync(clusterId)), [clusterId, dispatch]);
+};
 
-  const fetchCluster = React.useCallback(() => dispatch(fetchClusterAsync(clusterId)), [
-    clusterId,
-    dispatch,
-  ]);
+const useClusterPolling = (clusterId: string) => {
+  const { forceReload, uiState } = useSelector(selectCurrentClusterState);
+  const dispatch = useDispatch();
+  const fetchCluster = useFetchCluster(clusterId);
+
   React.useEffect(() => {
     if (forceReload) {
-      dispatch(forceReloadAction(false));
-      fetchCluster();
+      if (![ResourceUIState.LOADING, ResourceUIState.RELOADING].includes(uiState)) {
+        fetchCluster();
+      }
     }
-    const timmer = setTimeout(() => dispatch(forceReloadAction(true)), POLLING_INTERVAL);
-    return () => clearTimeout(timmer);
-  }, [fetchCluster, dispatch, forceReload]);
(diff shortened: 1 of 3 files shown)


jtomasek:
Is dispatch(forceReloadAction(true)) going to cancel previous setTimeout? I think there should be clearTimeout called at some point.

I was trying to put the logic into the fetchClusterAsync thunk action:
In that action
```
clearTimeout(timeout stored in redux)
await getClusterApiCall
const timeout = setTimeout(() => dispatch(fetchClusterAsync(clusterId), 5000)
return { data, timeout }
```
...

but I am getting into typescript hell :/

mareklibra:
Yeah, I wished to keep it simple.
The `clearTimeout()` handles just a corner case when the user repeatedly navigates out of the page (like to the cluster list) and comes back within a single polling interval. In that case, the count of useless requests can be up to the count of such steps.
Anyway, I have added such cleaning. Wdyt?

src/components/clusters/ClusterPage.tsx
```diff
@@ -7,27 +7,52 @@ import { ErrorState, LoadingState } from '../ui/uiState';
...
+      if (![ResourceUIState.LOADING, ResourceUIState.RELOADING].includes(uiState)) {
+        fetchCluster();
+      }
+    }
+    dispatch(forceReloadAction(false));
+  }, [fetchCluster, dispatch, forceReload, uiState]);
 
   React.useEffect(() => {
     fetchCluster();
+    const timmer = setInterval(() => dispatch(forceReloadAction(true)), POLLING_INTERVAL);
```
jtomasek, line 42: typo:
```suggestion
    const timer = setInterval(() => dispatch(forceReloadAction(true)), POLLING_INTERVAL);
```

src/components/clusters/ClusterPage.tsx
```diff
@@ -7,27 +7,52 @@ import { ErrorState, LoadingState } from '../ui/uiState';
...
     return () => {
+      clearInterval(timmer);
```
jtomasek, line 44:
```suggestion
      clearInterval(timer);
```

src/features/clusters/currentClusterSlice.ts
```diff
@@ -19,11 +19,13 @@ export const fetchClusterAsync = createAsyncThunk(
 type CurrentClusterStateSlice = {
   data?: Cluster;
   uiState: ResourceUIState;
+  forceReload: number;
```
jtomasek, line 22: how about renaming this to `isPolling`

src/features/clusters/currentClusterSlice.ts
```diff
@@ -32,12 +34,19 @@ export const currentClusterSlice = createSlice({
   reducers: {
     updateCluster: (state, action: PayloadAction<Cluster>) => ({ ...state, data: action.payload }),
     cleanCluster: () => initialState,
+    forceReload: (state, action: PayloadAction<boolean>) => ({
```
jtomasek, line 37: how about turning this into 2 actions: `startPolling` and `stopPolling`? It would IMO make the code more understandable

mareklibra: I can split into two actions but named `forceReload` and `cancelReload`. The original idea behind this PR was to provide an option how to "speed-up" polling from outside (like after an action not returning data update or similar).

src/features/clusters/currentClusterSlice.ts
mareklibra: `isReloadScheduled` ?

PR #73, opened 2020-04-30
Rearange host detail

mareklibra:
The host details are vertically split into 3 sections - meta, disks, nics.
Disks and nics are listed in a table.

Depends on:
- [x] #67


mareklibra:
details

PR #80, opened 2020-05-05
Add workflow triggered on Pull Requests to master

mareklibra:


mareklibra:
So far investigating if we can remove travis dependency

mareklibra:
Result: We can replace travis by GitHub worflows, they are triggered and notifications are sent. Works fine for our simple use case.

Benefit: single environment

Draw-back: slightly slower, travis CI finishes in a minute, GitHub requires multiple minutes.

I do not think we need something more complex than we have recently implemented except adding unit tests. Anyway, we can reuse code by using `actions` between different workflows.

@jtomasek , wdyt?

.github/workflows/pullrequest.yaml
```diff
@@ -0,0 +1,27 @@
...
+        uses: actions/checkout@v1
+      - name: Install dependencies
+        uses: borales/actions-yarn@v2.0.0
+        with:
+          cmd: install
+      - name: Call linter
+        uses: borales/actions-yarn@v2.0.0
+        with:
+          cmd: lint
+      - name: Build code
```
jtomasek, line 22: is build needed here?

mareklibra: I think so, kind of test. We will add extra step for tests once we have any.

PR #81, opened 2020-05-05
Remove extra polling of hosts

mareklibra:
Host list retrieved along cluster details is used instead.

In addition, manual passing down of parameters is removed in favour of Redux.

### DIFF
--- a/src/components/clusterWizard/BaremetalInventory.tsx
+++ b/src/components/clusterWizard/BaremetalInventory.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { useSelector } from 'react-redux';
 import {
   ButtonVariant,
   Breadcrumb,
@@ -8,7 +9,6 @@ import {
   Text,
   TextContent,
 } from '@patternfly/react-core';
-
 import PageSection from '../ui/PageSection';
 import HostsTable from './HostsTable';
 import ClusterWizardToolbar from './ClusterWizardToolbar';
@@ -16,10 +16,9 @@ import { ToolbarButton, ToolbarText } from '../ui/Toolbar';
 import { Cluster } from '../../api/types';
 import { Link } from 'react-router-dom';
 import { WizardStep } from '../../types/wizard';
-import { getClusterHosts } from '../../api/clusters';
-import useApi from '../../api/useApi';
 import { ResourceUIState } from '../../types';
 import { DiscoveryImageModalButton } from './discoveryImageModal';
+import { selectCurrentClusterUIState } from '../../selectors/currentCluster';
 
 interface BareMetalInventoryProps {
   cluster: Cluster;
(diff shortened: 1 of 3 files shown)


src/components/clusterWizard/ClusterWizardForm.tsx
```diff
@@ -17,42 +18,42 @@ import {
...
 
 interface ClusterWizardFormProps {
-  cluster: Cluster;
   setStep: React.Dispatch<React.SetStateAction<WizardStep>>;
 }
 
-const ClusterWizardForm: React.FC<ClusterWizardFormProps> = ({ cluster, setStep }) => {
+const ClusterWizardForm: React.FC<ClusterWizardFormProps> = ({ setStep }) => {
   const dispatch = useDispatch();
+  const cluster = useSelector(selectCurrentCluster);
```
jtomasek, line 43: Hmm, having to consider cluster being undefined is quite bad. We're already making sure that cluster is set on ClusterPage so we can be sure that cluster is always set in this component and also in ClusterWizard. I think a better solution is to keep passing the cluster as prop as it was before. Alternatively we can use `connect` but I think using props here is better.

src/components/clusterWizard/BaremetalInventory.tsx
```diff
@@ -8,36 +9,22 @@ import {
...
-  });
-
-  React.useEffect(() => {
-    const timer = setTimeout(() => {
-      fetchHosts();
-    }, 5000);
-    return () => clearTimeout(timer);
-  });
+const BaremetalInventory: React.FC<BareMetalInventoryProps> = ({ setStep }) => {
+  const { data: cluster, uiState } = useSelector(selectCurrentClusterState);
```
jtomasek, line 27: with keeping the cluster prop we can just get the uiState using `useSelector(selectCurrentClusterUIState)`

src/components/clusterWizard/BaremetalInventory.tsx
```diff
@@ -46,7 +33,7 @@ const BaremetalInventory: React.FC<BareMetalInventoryProps> = ({ cluster, setSte
           <BreadcrumbItem>
             <Link to="/clusters">Clusters</Link>
           </BreadcrumbItem>
-          <BreadcrumbItem isActive>{cluster.name}</BreadcrumbItem>
+          <BreadcrumbItem isActive>{cluster?.name || ''}</BreadcrumbItem>
```
jtomasek, line 36: Cluster always exists otherwise this component won't ever get rendered.

src/components/clusterWizard/BaremetalInventory.tsx
```diff
@@ -59,12 +46,7 @@ const BaremetalInventory: React.FC<BareMetalInventoryProps> = ({ cluster, setSte
         </TextContent>
       </PageSection>
       <PageSection variant={PageSectionVariants.light} isMain>
-        <HostsTable
-          hosts={hosts || cluster.hosts}
-          uiState={uiState}
-          fetchHosts={fetchHosts}
-          clusterId={cluster.id}
-        />
+        {cluster && <HostsTable cluster={cluster} uiState={uiState} />}
```
jtomasek, line 49: same as above

src/components/clusterWizard/ClusterWizardForm.tsx
mareklibra: You are right, my intention was to make the flow simpler but actually it not become truth, I am removing the last commit.

PR #85, opened 2020-05-05
Introduce NewClusterModal

jtomasek:
This allows setting cluster name and openshiftVersion prior to cluster creation

Screenshot 2020-05-05 at 14 15 25


src/components/clusters/Clusters.tsx
```diff
@@ -83,12 +83,7 @@ const Clusters: React.FC = () => {
             <ClustersTable rows={clusterRows} deleteCluster={deleteCluster} />
           </PageSection>
           <ClusterWizardToolbar>
-            <ToolbarButton
-              variant={ButtonVariant.primary}
-              component={(props) => <Link to="/clusters/new" {...props} />}
-            >
-              Create New Cluster
-            </ToolbarButton>
+            <NewClusterModalButton ButtonComponent={ToolbarButton} />
```
mareklibra, line 86: Why do we need this flexibility with button component?

mareklibra:
We should not allow the user to create a new cluster with conflicting name (of an existing cluster). Opening a jira ticket for that.

src/components/clusters/Clusters.tsx
jtomasek: You're right, it is not needed here in this case (I've copied it over from discovery Iso modal which is triggered from 2 different buttons.

jtomasek:
> We should not allow the user to create a new cluster with conflicting name (of an existing cluster).

I'll fix this as part of https://github.com/openshift-metal3/facet/pull/87

PR #88, opened 2020-05-05
Added tool to generate the deployment file from template

ronniel1:
Added a tool to generate the deployment file from the template.
This tool is copied to the image.
Anyone needing to deploy can run the tool directly from the image:
```    
docker run --rm quay.io/ocpmetal/ocp-metal-ui:latest deploy/deploy_config.sh
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,9 @@
 .PHONY: all build deploy clean
 
 IMAGE := $(or ${IMAGE},quay.io/ocpmetal/ocp-metal-ui:latest)
+BM_INVENTORY_URL := $(or ${BM_INVENTORY_URL},http://bm-inventory.default.svc.cluster.local:8090)
+
+DEPLOY_FILE="deploy/ocp-metal-ui.yaml"
 
 all: build deploy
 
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -10,9 +10,10 @@ build:
 	sudo podman push $(IMAGE)
 
 deploy:
-	sed "s#__IMAGE__#${IMAGE}#g;s#__BM_INVENTORY_URL__#${BM_INVENTORY_URL}#g" deploy/ocp-metal-ui-template.yaml > deploy/ocp-metal-ui.yaml
-	kubectl apply -f deploy/ocp-metal-ui.yaml
-	rm deploy/ocp-metal-ui.yaml
+	mkdir -p build/deploy/
+	tools/deploy_config.sh "${BM_INVENTORY_URL}" "${IMAGE}" deploy/ocp-metal-ui-template.yaml > build/deploy/ocp-metal-ui.yaml
```
jtomasek, line 14: Shouldn't the output go to `deploy/ocp-metal-ui.yaml` rather than `build/deploy/...`?
```suggestion
	tools/deploy_config.sh "${BM_INVENTORY_URL}" "${IMAGE}" deploy/ocp-metal-ui-template.yaml > deploy/ocp-metal-ui.yaml
```

jtomasek: The build directory should contain only built static files which are then served from the nginx server. We should not include these files there.

Makefile
```diff
@@ -10,9 +10,10 @@ build:
...
+	kubectl apply -f build/deploy/ocp-metal-ui.yaml
```
jtomasek, line 15: Why not just deleting the generated file here?

tools/deploy_config.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
jtomasek, line 1: It would be nice to keep everything related to deploying in `/deploy` dir. Can we move this to `deploy/deploy_config`?

Makefile
ronniel1: It is usually good practice to only modify files under the build/ dir in the build process.

Makefile
ronniel1: see above
Also this helps in debugging deploy errors.
Otherwise we do not know what the deploy file contained

tools/deploy_config.sh
ronniel1: sure

Makefile
jtomasek: good point, doing it in `clean` is good enough

PR #99, opened 2020-05-12
Rename vCPU to CPU, split cores

mareklibra:


mareklibra:
Need to clarify exact semantics of the `hwInfo.cpu.cpus` field? Is it total-cores-per-machine? Or cores per single CPU chip?

PR #100, opened 2020-05-12
HostsTable is always compact-expandable

mareklibra:


mareklibra:
This patch is inlcuded in https://github.com/openshift-metal3/facet/pull/100

PR #102, opened 2020-05-12
Change size of Host Detail headings

mareklibra:


mareklibra:
redesignHostDetail

mareklibra:
I am considering changing the Host Detail section to use [Patternfly's Data list](https://www.patternfly.org/v4/documentation/react/components/text#data-list).

@andybraren, wdyt ?

andybraren:
@mareklibra A description list (I think the documentation is mis-labeled?) sounds good structurally with `<dl>`, `<dt>`, and `<dd>`, but I notice that both OCM and the OCP Web Console avoid using the `.pf-c-content` class's styling in order to make the title/value pairs vertical instead of horizontal, like this:

image

image

We should probably align with the vertical orientation of OCM/OCP here, but I wonder why a vertical orientation isn't natively supported by the component. Maybe there's a different component we should be using? WDYT @matthewcarleton?

matthewcarleton:
@andybraren we had a similar issue and ended up just rolling our own b/c we wanted the horizontal layout. I'm not sure why it isn't, probably just hasn't seen the request come in from the product side. A good reason to create an issue :)

PR #108, opened 2020-05-13
Deploy + Progress

jtomasek:
Initial work for starting the installation and tracking the progress.
Configuration validation and proper progress tracking will be done as
a follow up work

### DIFF
--- a/src/components/clusterConfiguration/HostsTable.tsx
+++ b/src/components/clusterConfiguration/HostsTable.tsx
@@ -59,6 +59,11 @@ const hostToHostTableRow = (openRows: OpenRows) => (host: Host, idx: number): IR
   const { id, status, statusInfo, role, createdAt, hardwareInfo = '' } = host;
   const hwInfo = getHardwareInfo(hardwareInfo) || {};
   const { cores, memory, disk } = getHostRowHardwareInfo(hwInfo);
+  const roleCellTitle = ['installing', 'installed', 'error'].includes(status) ? (
+    role
+  ) : (
+    <RoleDropdown role={role} host={host} />
+  );
 
   return [
     {
@@ -67,8 +72,8 @@ const hostToHostTableRow = (openRows: OpenRows) => (host: Host, idx: number): IR
       cells: [
         id,
         {
-          title: <RoleDropdown role={role} host={host} />,
-          sortableValue: host.role,
+          title: roleCellTitle,
+          sortableValue: role,
         },
         id, // TODO(mlibra): should be serial number
         {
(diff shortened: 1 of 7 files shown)


src/components/clusterConfiguration/HostsTable.tsx
```diff
@@ -59,7 +64,7 @@ const hostToHostTableRow = (openRows: OpenRows) => (host: Host, idx: number): IR
       cells: [
         id,
         {
-          title: <RoleDropdown role={role} host={host} />,
+          title: roleCellTitle,
```
mareklibra, line 67: I would disable change via props within `RoleDropdown` component instead of changing it to static text

src/components/clusterProgress/ClusterProgress.tsx
```diff
@@ -0,0 +1,67 @@
...
+          </BreadcrumbItem>
+          <BreadcrumbItem isActive>{cluster.name}</BreadcrumbItem>
+        </Breadcrumb>
+      </PageSection>
+      <PageSection variant={PageSectionVariants.light} isMain>
+        <TextContent>
+          <Text component="h1">{cluster.name}</Text>
+          <dl>
+            <dt>Creation started at</dt>
+            <dd>{new Date(Date.parse(cluster.installStartedAt || '')).toLocaleString()}</dd>
```
mareklibra, line 38: Can we move this to a helper function?
In addition, can we use constructor for parsing the input string?

src/components/ui/formik/validationSchemas.ts
```diff
@@ -19,9 +19,7 @@ export const getUniqueNameValidationSchema = (excludedList: (string | undefined)
     (value) => !excludedList.includes(value),
   );
 
-export const sshPublicKeyValidationSchema = Yup.string()
-  .matches(SSH_PUBLIC_KEY_REGEX, {
-    message: 'SSH public key must consist of "ssh-rsa key [email]"',
-    excludeEmptyString: true,
-  })
-  .required('Required');
+export const sshPublicKeyValidationSchema = Yup.string().matches(SSH_PUBLIC_KEY_REGEX, {
```
mareklibra, line 22: nit: is it result of prettier? Did not complain to me and I like former style more.

jtomasek: yeah, prettier formats it this way

src/components/clusterConfiguration/HostsTable.tsx
jtomasek: Hmm, that would imply that there is a certain possibility to eventually change the role but after installation is started it is always fixed

src/components/clusterProgress/ClusterProgress.tsx
jtomasek: Done.

src/components/clusterConfiguration/HostsTable.tsx
mareklibra: We can move responsibility for the whole cell down to the component. But ok, it's nitpicking.

PR #110, opened 2020-05-14
ClustersTable: add sorting

mareklibra:
Depends on:
- [x] #109

### DIFF
--- a/src/api/utils.ts
+++ b/src/api/utils.ts
@@ -1,12 +1,6 @@
 import Axios, { AxiosError } from 'axios';
-import { ISortBy, SortByDirection, IRow } from '@patternfly/react-table';
 
 type OnError = <T>(arg0: AxiosError<T>) => void;
-
-export type HumanizedSortable = {
-  title: string | React.ReactNode;
-  sortableValue: number | string;
-};
 
 export const handleApiError = <T>(error: AxiosError<T>, onError?: OnError) => {
   if (Axios.isCancel(error)) {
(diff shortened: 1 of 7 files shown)


src/components/clusters/ClustersTable.tsx
```diff
@@ -10,44 +21,80 @@ interface ClustersTableProps {
   deleteCluster: (id: string) => void;
 }
 
+// TODO(mlibra): let's try to remove it - not needed for HostsTable
```
mareklibra, line 24: I will remove this extra formatting in a follow-up and let's discuss there.

mareklibra: https://github.com/openshift-metal3/facet/pull/111

src/api/utils.ts
```diff
@@ -25,3 +31,25 @@ export const handleApiError = <T>(error: AxiosError<T>, onError?: OnError) => {
     if (onError) return onError(error);
   }
 };
+
+type getCellType = (row: IRow, index: number | undefined) => string | HumanizedSortable;
+
+export const rowSorter = (sortBy: ISortBy, getCell: getCellType) => (a: IRow, b: IRow): number => {
```
jtomasek, line 37: Would it make more sense to move this into `components/ui/table/utils.ts`?

src/components/clusters/ClustersTable.tsx
```diff
@@ -1,7 +1,18 @@
 import React from 'react';
-import { Table, TableHeader, TableBody, TableVariant, IRowData } from '@patternfly/react-table';
+import {
+  Table,
+  TableHeader,
+  TableBody,
+  IRowData,
+  sortable,
```
jtomasek, line 7: this needs to be used from `src/components/ui/table/sortable`

src/api/utils.ts
mareklibra: done

src/components/clusters/ClustersTable.tsx
mareklibra: done

src/components/clusterConfiguration/HostRowDetail.tsx
```diff
@@ -32,7 +32,8 @@ type HostDetailItemProps = {
         value?: string;
       }[]
     | string
-    | number;
+    | number
+    | React.ReactNode;
```
jtomasek, line 36: `ReactNode` already includes `string` and `number`

PR #111, opened 2020-05-14
ClusterTable: remove headerConfig

mareklibra:
We don't need this tweak anymore.

Depends on:
- [x] #110


mareklibra:
With:
01_with

Without:
02_without

PR #112, opened 2020-05-14
Add validations to ClusterConfiguration page

mareklibra:


mareklibra:
> None of the properties should be required except for name IMHO. Eventually we'll add another validation when triggering the installation - at that point the properties should be required.

Removed.

PR #114, opened 2020-05-14
Fix disappearing new cluster modal in empty state

rawagner:
The issue is caused by `RELOADING` ResourceUIState which causes[ switch statement](https://github.com/openshift-metal3/facet/blob/master/src/components/clusters/Clusters.tsx#L56) to use the `default` path.

### DIFF
--- a/src/components/clusters/Clusters.tsx
+++ b/src/components/clusters/Clusters.tsx
@@ -16,11 +16,17 @@ import { AddCircleOIcon } from '@patternfly/react-icons';
 import { ResourceUIState } from '../../types';
 import ClustersTable from './ClustersTable';
 import { deleteClusterAsync, fetchClustersAsync } from '../../features/clusters/clustersSlice';
-import { NewClusterModalButton } from './newClusterModal';
+import { NewClusterModalButton, NewClusterModal } from './newClusterModal';
 
 const Clusters: React.FC = () => {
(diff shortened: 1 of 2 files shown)


src/components/clusters/Clusters.tsx
```diff
@@ -16,11 +16,17 @@ import { AddCircleOIcon } from '@patternfly/react-icons';
...
 
 const Clusters: React.FC = () => {
+  const { LOADING, EMPTY, ERROR, RELOADING } = ResourceUIState;
+  const [isModalOpen, setIsModalOpen] = React.useState(false);
   const clusterRows = useSelector(selectClusterTableRows);
   const clustersUIState = useSelector(selectClustersUIState);
+  const uiState = React.useRef(clustersUIState);
+  if (clustersUIState !== RELOADING) {
+    uiState.current = clustersUIState;
+  }
```
jtomasek, line 29: With a slight change to the `clustersSlice` this is not going to be needed:

replace https://github.com/openshift-metal3/facet/blob/master/src/features/clusters/clustersSlice.ts#L43
with
`const uiState = state.uiState === LOADED && state.data.length ? RELOADING : LOADING;`

rawagner: That would cause a Loading state to appear if you start from empty state -> create new cluster -> cancel

PR #115, opened 2020-05-14
Add help tooltip to form fields

rawagner:
Screenshot from 2020-05-14 16-30-00

### DIFF
--- a/src/components/ui/formik/HelpTooltip.tsx
+++ b/src/components/ui/formik/HelpTooltip.tsx
@@ -1,0 +1,20 @@
+import * as React from 'react';
+import { Tooltip } from '@patternfly/react-core';
+import { OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
+
+import './help-tooltip.scss';
+
+const HelpTooltip: React.FC<HelpTooltipProps> = ({ helperText }) =>
+  helperText ? (
+    <span className="help-tooltip">
+      <Tooltip content={helperText} tippyProps={{ interactive: true }}>
+        <OutlinedQuestionCircleIcon />
+      </Tooltip>
+    </span>
+  ) : null;
+
+export default HelpTooltip;
+
+type HelpTooltipProps = {
+  helperText: React.ReactNode;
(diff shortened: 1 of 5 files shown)


jtomasek:
Aren't we breaking the patternfly usage guidelines by doing this? https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/forms#form-writing-guidance (_Never hide critical information in a popover._)

IMHO we should stick to patternfly guidelines rather than trying to align with ACM. It should be the other way around.

We should identify critical information and keep it in the help-block and put only less important additional info into the popups when needed. Popover is quite hard to use/copy text from etc.

andybraren:
A few notes:

1. [PatternFly's form design guideline](https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/forms#popovers) has an image of a white popover component being used to display the "?" text instead of a dark tooltip, but I can't seem to find an example of it in [the React docs](https://www.patternfly.org/v4/documentation/react/components/form). Is that an easy swap? That might help with the copying issue. I dislike tooltips in general partly for that reason. 🤷‍♂️

2. You're right @jtomasek, there's some critical info that should be in [helper text](https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/forms#helper-text) below the form fields instead of within the "?" icon. The slightly simplified field info icon text I proposed in the [Installer design doc](https://docs.google.com/document/d/1OtAqrwa8dZYFp1jF8oBZ72wb3OqmjAgqvTgckXTE_qo/edit#heading=h.aaruszpmj2uw) is a little better, but there's still the problem of:

3. There seem to be multiple field label, popover text, and help text inconsistencies across ACM, OCM (for Dedicated clusters), and now this installer UI that need to be worked through. I'll reach out to the UXDers in those teams to see if we can align all of them and figure out what critical info should always be shown as help text, what goes in the "?" icon, and what the field labels should be for each one.

mareklibra:
@rawagner, is this still valid?

rawagner:
probably not, closing

PR #118, opened 2020-05-19
Collapse HostsTable rows when sorting

mareklibra:


jtomasek:
It seems the error still persists.
Screenshot 2020-05-20 at 12 49 04

Steps to reproduce:
1. click 'Role' column title to switch the ordering
2. expand first host

Result (screenshot): Second host gets expanded instead of the first one.

mareklibra:
Rebased only

mareklibra:
@jtomasek , fixed

PR #122, opened 2020-05-20
Introduce cluster progress tracking and cluster detail logic

jtomasek:
- Cluster installation progress and detail are handled by ClusterDetail component
- Depending on the cluster status the ClusterProgress, ClusterCredentials and
  ClusterInstallationError components are rendered
- Proper cluster tracking depends on progessInfo property on the cluster which is
  not yet available, so it is worked around in this patch. A proper integration
  will be done as a follow up once progressInfo is available
- Follow ups:
  - per host progress tracking
  - Make ClusterCredentials component fetch real data
  - Implement 'Abort installation' action


src/components/clusterDetail/ClusterCredentials.tsx
```diff
@@ -0,0 +1,31 @@
...
+  // TODO(jtomasek): Fetch cluster credentials data
+  // TODO(jtomasek):
+  return (
+    <GridItem>
+      <TextContent>
+        <dl className="cluster-detail__details-list">
+          <dt>Web Console URL</dt>
+          <dd>https://console-openshift-console.apps.clusterName.dev.domain.com</dd>
+          <dt>Username</dt>
+          <dd>kubeadmin</dd>
```
mareklibra, line 19: Leftover?

src/components/clusterDetail/ClusterDetail.tsx
```diff
@@ -0,0 +1,106 @@
...
+import HostsTable from '../clusterConfiguration/HostsTable';
+import ClusterToolbar from '../clusters/ClusterToolbar';
+import { ToolbarButton } from '../ui/Toolbar';
+import { getHumanizedDateTime } from '../ui/utils';
+import ClusterBreadcrumbs from '../clusters/ClusterBreadcrumbs';
+import ClusterProgress from './ClusterProgress';
+import ClusterCredentials from './ClusterCredentials';
+
+import './clusterDetail.css';
+import ClusterInstallationError from './ClusterInstallationError';
```
mareklibra, line 22: nit: css at last

src/components/clusters/ClusterStatus.tsx
```diff
@@ -0,0 +1,51 @@
...
+  WarningTriangleIcon,
+  CheckCircleIcon,
+  UnknownIcon,
+  InProgressIcon,
+} from '@patternfly/react-icons';
+import { Cluster } from '../../api/types';
+import { CLUSTER_STATUS_LABELS } from '../../config/constants';
+
+const getStatusIcon = (status: Cluster['status']) => {
+  if (status === 'insufficient') return <WarningTriangleIcon color={warningColor.value} />;
```
mareklibra, line 19: What about `switch` without `default` so linter can warn us to add a section when the type changes?

src/components/clusters/ClusterStatus.tsx
```diff
@@ -0,0 +1,51 @@
...
+  status: Cluster['status'];
+  statusInfo?: string;
+};
+
+const ClusterStatus: React.FC<ClusterStatusProps> = ({ status, statusInfo }) => {
+  const title = CLUSTER_STATUS_LABELS[status];
+  const icon = getStatusIcon(status);
+  if (statusInfo) {
+    return (
+      <Popover headerContent={<div>{title}</div>} bodyContent={<div>{statusInfo}</div>}>
```
mareklibra, line 37: These `<div>`s will force re-rendering in every cycle.

src/config/constants.ts
```diff
@@ -19,3 +19,11 @@ export const POLLING_INTERVAL = 10000;
 export const HOST_ROLES = ['worker', 'master'];
 // Without undefined. Otherwise must conform generated Host['roles'] - see api/types.ts
 export type HostRolesType = 'master' | 'worker';
+
+export const CLUSTER_STATUS_LABELS = {
```
mareklibra, line 23: what about adding type for the `key` to have better guarding?

src/components/clusterDetail/ClusterCredentials.tsx
jtomasek: It is rather a placeholder, should I remove it? I don't think this gets rendered as we're not able to get to 'installed' state yet.

src/components/clusterDetail/ClusterDetail.tsx
jtomasek: Done

src/components/clusters/ClusterStatus.tsx
jtomasek: Done

src/config/constants.ts
jtomasek: Done

PR #126, opened 2020-05-22
Introduce facet-lib

mareklibra:
In attempt to move all reusable components and related code to a library.

Homepage: https://github.com/mareklibra/facet-lib

To be done in follow-ups:
- reexport selected components
- setup CI on facet-lib
- move facet-lib under same repo as the `facet` project (which is about to be moved too)
- test reusability in OCM and OCP projects


mareklibra:
Deprecated by #218, closing

PR #128, opened 2020-05-25
Add delete modal

rawagner:
Screenshot from 2020-05-25 11-36-16


rawagner:
I still need to implement tests

src/components/clusters/ClustersTable.tsx
```diff
@@ -38,20 +39,22 @@ const columns = [
...
+  });
 
   const actions = React.useMemo(
     () => [
       {
         title: 'Delete',
-        onClick: (event: React.MouseEvent, rowIndex: number, rowData: IRowData) =>
-          deleteCluster(rowData.id.title),
+        onClick: (event: React.MouseEvent, rowIndex: number, rowData: IRowData) => {
+          setDeleteClusterRow(rowData);
```
mareklibra, line 53: What is the benefit of storing whole row here? I would prefer just identification of the cluster to be deleted.

rawagner: Storing only name and id instead of the whole row now

PR #132, opened 2020-05-26
Add container image with E2E tests

mareklibra:


hacks/run-tests.sh
```diff
@@ -0,0 +1,23 @@
...
+
+echo Connecting to UI at: ${CYPRESS_BASE_URL}
+echo Test image: ${TEST_IMAGE}
+
+VIDEO_DIR=${BASE_DIR}/videos
+SCREENSHOT_DIR=${BASE_DIR}/screenshots
+
+mkdir -p ${VIDEO_DIR}
+mkdir -p ${SCREENSHOT_DIR}
+docker run -it \
```
jtomasek, line 15: We should use either docker or podman across our scripts (Makefile uses podman).

jtomasek:
How is the run_tests.sh going to be used? Is it going to be also part of the image so that the test image is the only thing needed to run the tests? (apart from running application of course).

jtomasek:
I'd imagine running the tests in similar manner as the UI is deployed - using the image rather than needing to pull UI codebase. https://github.com/tsorya/test-infra/blob/master/scripts/deploy_ui.sh#L17

hacks/run-tests.sh
mareklibra: done

mareklibra:
Actually I am thinking about this script as an equivalent of the single https://github.com/tsorya/test-infra/blob/master/scripts/deploy_ui.sh#L17 line. In fact, it's just about setting `podman` command-line to be provided with URL and mounts to local (out-of container) storage.

The use case is recently considered to be
- by test-infra
- developer on laptop

So I can include similar script in the container to avoid dependency on downloading the sources/script from github. It means the execution will consist of two commands:
```
$ podman cp [script_in_container] run-tests.sh]
$ CYPRESS_BASE_URL=[URL] sh ./run-tests.sh
```

The script stored within the container can contain exact tag (not just `:latest`). This is just a default which can be overridden by `TEST_IMAGE` env prop.

README.md will be updated accordingly.

@jtomasek , wdyt?

mareklibra:
I have added commit implementing this change but honestly, **I think this is overcomplicating things.** As written above, the former script was just about setting single `podman run` command.

The tests-image contains `/run-tests.sh` which is meant to be executed outside of the container to parametrize execution of this container. This `run-tests.sh` is altered for the container tag while building it.

There is newly `hacks/run-tests-latest.sh` simplifying geting the `run-tests.sh` out of the container, so to run the tests the user needs:
```
export CYPRESS_BASE_URL=http://localhost:3000
  curl https://raw.githubusercontent.com/openshift-metal3/facet/master/hacks/run-tests-latest.sh | sh
```

mareklibra:
I have reverted last commit - it would be really over-complication without clear benefits.
The `test-infra` can mimic `hacks/run-tests.sh`, there are just two commands - `mkdir` and `podman run`.

PR #135, opened 2020-05-26
Update host progress to match current state of the API

jtomasek:
- remove unimplemented cluster action buttons


src/config/constants.ts
```diff
@@ -35,6 +35,7 @@ export const HOST_STATUS_LABELS: { [key in Host['status']]: string } = {
   insufficient: 'Insufficient',
   disabled: 'Disabled',
   installing: 'Installing',
+  'installing-in-progress': 'Installing',
```
mareklibra, line 38: Don't we need to differentiate these two states? How long is normally expected to be in `installing` state (not in `installing-in-progress`). What about `Installation pending`?

gamli75: When in status Installing - get next step will return the install command
When in status installing-in-progress - it means that the installation process was started and we updated the installation progress - in this state get-next-step will not return any command.
we added this state in order to avoid starting the installation twice.

jtomasek: I've as agreed in the Slack discussion, I've updated label for 'installing' to be 'Starting Installation'

PR #137, opened 2020-05-26
Add kubeconfig download

rawagner:


src/components/clusterDetail/ClusterCredentials.tsx
```diff
@@ -21,7 +24,10 @@ const ClusterCredentials: React.FC<ClusterCredentialsProps> = () => {
           <dd>-------------</dd>
         </dl>
       </TextContent>
-      <Button variant={ButtonVariant.secondary} isDisabled>
+      <Button
+        variant={ButtonVariant.secondary}
+        onClick={() => saveAs(getClusterFileURL(cluster.id, 'kubeconfig'))}
```
jtomasek, line 29: It would be nice to add error handling here, but we can add it as a subsequent change as part of the error handling work.

PR #140, opened 2020-05-27
Credentials

rawagner:
Screenshot from 2020-05-27 12-47-52


src/api/types.ts
```diff
@@ -432,3 +432,9 @@ export interface SystemVendor {
   productName?: string;
   manufacturer?: string;
 }
+
```
jtomasek, line 435: this file is autogenerated, put it somewhere else for the time being please

src/components/clusterDetail/ClusterCredentials.tsx
```diff
@@ -1,35 +1,81 @@
...
+          <dd>{credentials.username}</dd>
           <dt>Password</dt>
-          <dd>-------------</dd>
+          <dd>
+            <ClipboardCopy
+              className="cluster-detail__password"
+              isReadOnly
+              onCopy={(event) => clipboardCopyFunc(event, credentials.password)}
+            >
+              &bull;&bull;&bull;&bull;&bull;
```
jtomasek, line 60: It's a bit pity that  ClipboardCopy won't allow us to pass the input type (password) it should render. :/

src/components/clusterDetail/ClusterDetail.tsx
```diff
@@ -36,6 +37,28 @@ const installationSteps = [
...
+        const response = await getClusterCredentials(cluster.id);
+        setCredentials(response.data);
+      } catch (err) {
+        setCredentialsError(err);
+      }
+    };
+    fetch();
+  }, [cluster.id]);
+
+  React.useEffect(() => {
```
jtomasek, line 56: We could avoid doing this by moving it into the ClusterCredentials. Is it on purpose to make the ClusterCredentials without side effects?

src/components/clusterDetail/ClusterCredentials.tsx
rawagner: at least you cant tell how long the password is :D

src/components/clusterDetail/ClusterDetail.tsx
rawagner: we still need console URL so fetching the data in this component makes sense. useEffect without side effects is completely fine IMO

src/components/clusterDetail/ClusterCredentials.tsx
rawagner: note: we also dont have `reveal password` button, but i would argue how useful that actually is

src/api/types.ts
rawagner: moved away for now

rawagner:
@jtomasek done

src/components/clusterDetail/ClusterCredentials.tsx
```diff
@@ -1,35 +1,81 @@
...
-        variant={ButtonVariant.secondary}
-        onClick={() => saveAs(getClusterFileURL(cluster.id, 'kubeconfig'))}
-      >
-        Download kubeconfig
-      </Button>
+    );
+  }
+
+  return (
+    <GridItem>
```
jtomasek, line 69:
```suggestion
    <GridItem span={12} lg={10} xl={6}>
```

Could solve the width problem without having to use custom css.

src/components/clusterDetail/ClusterDetail.css
```diff
@@ -5,3 +5,7 @@
 .cluster-detail__details-list dd {
   margin-bottom: var(--pf-global--spacer--sm);
 }
+
+.cluster-detail__password {
```
jtomasek, line 9: not needed with suggestion above

src/components/clusterDetail/ClusterCredentials.tsx
rawagner: fixed

PR #144, opened 2020-05-28
Enhance error handling to include API error message

jtomasek:
- Convert AlertsSection to be universal way to display page erorrs
- Add error handling for deleting cluster
- Update form errors to include API based error messages
- List of page alerts state is managed per page using useReducer

### DIFF
--- a/src/features/alerts/alertsSlice.ts
+++ b/src/features/alerts/alertsSlice.ts
@@ -1,0 +1,34 @@
+import { v4 as uuidv4 } from 'uuid';
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+import { AlertVariant } from '@patternfly/react-core';
+
+type AlertPayload = {
+  title: string;
+  message?: string;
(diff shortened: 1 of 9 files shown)


jtomasek:
We need to make a decision whether to keep the alerts in redux and persist them across pages (as this PR does) or if we track the alerts per page and discard them on page change (stored in page component state).

src/features/alerts/alertsSlice.ts
```diff
@@ -0,0 +1,34 @@
...
+};
+
+const initialState: AlertProps[] = [];
+
+export const alertsSlice = createSlice({
+  initialState,
+  name: 'alerts',
+  reducers: {
+    addAlert: (state, action: PayloadAction<AlertPayload>) => [
+      { key: uuidv4(), variant: AlertVariant.danger, ...action.payload },
```
mareklibra, line 25: Nice.
nit: `key` ==> `id`

PR #149, opened 2020-05-28
Whitelist host states when host role selection is enabled

jtomasek:


src/components/clusterConfiguration/HostsTable.tsx
```diff
@@ -54,12 +54,16 @@ const hostToHostTableRow = (openRows: OpenRows) => (host: Host): IRow => {
   const { id, status, statusInfo, role, createdAt, hardwareInfo = '' } = host;
   const hwInfo = getHardwareInfo(hardwareInfo) || {};
   const { cores, memory, disk } = getHostRowHardwareInfo(hwInfo);
-  const roleCellTitle = ['installing', 'installing-in-progress', 'installed', 'error'].includes(
-    status,
-  ) ? (
-    role
-  ) : (
+  const roleCellTitle = [
```
mareklibra, line 57: nit: I think this logic can be inside the `RoleDropdown` (or renamed component) which is specific for this use case.

Next time, when we are adding more status-dependent logic, we should introduce functions wrapping this conditional logic (like `isInstalling()`) and keep them co-located for better future maintenance.

jtomasek: I've moved the logic into a new `RoleCel`' component. I've also moved all host-related components and utils into `hosts/` directory.

src/components/hosts/RoleCell.tsx
```diff
@@ -0,0 +1,14 @@
...
+import { RoleDropdown } from './RoleDropdown';
+import { Host } from '../../api/types';
+import { canEditRole } from './utils';
+
+type RoleCellProps = {
+  host: Host;
+};
+
+const RoleCell: React.FC<RoleCellProps> = ({ host }) => {
+  return canEditRole ? <RoleDropdown host={host} /> : <>{host.role}</>;
```
mareklibra, line 11: Missing calling of the function :-)

jtomasek: fixed

PR #151, opened 2020-05-29
Adapt UI to the new Host Inventory

mareklibra:
Former hardwareInfo (Introspection) is not used anymore.

### DIFF
--- a/src/components/hosts/HostsTable.tsx
+++ b/src/components/hosts/HostsTable.tsx
@@ -18,7 +18,7 @@ import { ExtraParamsType } from '@patternfly/react-table/dist/js/components/Tabl
 import { AlertVariant } from '@patternfly/react-core';
 import { EmptyState } from '../ui/uiState';
 import { getColSpanRow, rowSorter } from '../ui/table/utils';
-import { Host, Cluster, Introspection } from '../../api/types';
+import { Host, Cluster, Inventory } from '../../api/types';
 import { enableClusterHost, disableClusterHost } from '../../api/clusters';
 import { Alerts, Alert } from '../ui/Alerts';
 import { getHostRowHardwareInfo, getHumanizedTime } from './hardwareInfo';
@@ -28,9 +28,9 @@ import { HostDetail } from './HostRowDetail';
 import { forceReload } from '../../features/clusters/currentClusterSlice';
 import { handleApiError, stringToJSON } from '../../api/utils';
 import sortable from '../ui/table/sortable';
+import RoleCell from './RoleCell';
 
 import './HostsTable.css';
-import RoleCell from './RoleCell';
 
 type HostsTableProps = {
   cluster: Cluster;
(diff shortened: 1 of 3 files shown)


mareklibra:
---

mareklibra:
@ori-amizur , can we get more data even from a virtual machine? 
- serial number can be libvirt's identifier or similar
- default value for speed should be 0

Is the `0.0.0.0` BMC address a hardcoded default? Can it be empty in that case?

ori-amizur:
> @ori-amizur , can we get more data even from a virtual machine?
> 
> * serial number can be libvirt's identifier or similar
> * default value for speed should be 0
> 
> Is the `0.0.0.0` BMC address a hardcoded default? Can it be empty in that case?

@mareklibra We can provide the system uuid. (dmidecode -s system-uuid). In case of a virtual machine it is the same as the host-id  Wouldn't it be misleading if we put it in the place of serial number?
The returned value of the nic speed is provided from /sys/class/net/<nic-name>/speed without interpretation.  We can change the value to 0 in case it is -1.
The 0.0.0.0 is hardcoded default.  It is the same as provided by ironic python agent.  It can be changed to empty string.

mareklibra:
Rebased

src/components/clusterConfiguration/HostsTable.tsx
```diff
@@ -41,7 +41,7 @@ type OpenRows = {
 };
 
 const columns = [
-  { title: 'Serial Number', transforms: [sortable], cellFormatters: [expandable] },
+  { title: 'Host name', transforms: [sortable], cellFormatters: [expandable] },
```
jtomasek, line 44: I think this should be called 'Hostname'

src/components/clusterConfiguration/HostsTable.tsx
```diff
@@ -67,7 +68,7 @@ const hostToHostTableRow = (openRows: OpenRows) => (host: Host): IRow => {
       // visible row
       isOpen: !!openRows[id],
       cells: [
-        id,
+        inventory.hostname,
```
jtomasek, line 71: The hostname is not set until the host gets discovered.

```suggestion
        inventory.hostname || EMPTY,
```

mareklibra: done

src/components/clusterConfiguration/HostsTable.tsx
mareklibra: done

src/components/hosts/HostsTable.tsx
```diff
@@ -51,16 +52,16 @@ const columns = [
...
+  const inventory = stringToJSON<Inventory>(inventoryString) || {};
+  const { cores, memory, disk } = getHostRowHardwareInfo(inventory);
 
   return [
     {
       // visible row
       isOpen: !!openRows[id],
       cells: [
-        id,
+        inventory.hostname || EMPTY,
```
jtomasek, line 64: Sorry for not noticing it earlier, I think the EMPTY needs to be similar to EMPTY on hw columns?
```suggestion
        inventory.hostname || { title: DASH, sortableValue: '' },
```

PR #153, opened 2020-06-01
Add specific validation for starting cluster installation

jtomasek:
- update NewClusterModal and ClusterConfiguration to run unique cluster name
  validation asynchronously on form submit


src/components/clusterConfiguration/ClusterConfiguration.tsx
```diff
@@ -59,13 +57,34 @@ type SubnetHostMap = {
...
+  clusterNetworkCidr: ipBlockValidationSchema,
+  serviceNetworkCidr: ipBlockValidationSchema,
+  apiVip: ipValidationSchema,
+  ingressVip: ipValidationSchema,
+  pullSecret: validJSONSchema,
+  sshPublicKey: sshPublicKeyValidationSchema,
+});
+
+const installValidationSchema = Yup.object().shape({
+  name: nameValidationSchema,
```
mareklibra, line 78: nit: what about generating these schemas from a shared object? Will simplify maintenance.

src/components/clusterConfiguration/ClusterConfiguration.tsx
```diff
@@ -143,31 +166,27 @@ const ClusterConfiguration: React.FC<ClusterConfigurationProps> = ({ cluster })
...
-        ingressVip: ipValidationSchema,
-        pullSecret: validJSONSchema,
-        sshPublicKey: sshPublicKeyValidationSchema,
-      }),
-    [clusterNames],
-  );
-
-  const handleSubmit = async (values: ClusterConfigurationValues) => {
-    const { submitType, ...params } = values;
+  const handleSubmit = async (
```
mareklibra, line 172: Should be `React.useCallback()`

src/components/clusterConfiguration/ClusterConfiguration.tsx
```diff
@@ -207,8 +230,10 @@ const ClusterConfiguration: React.FC<ClusterConfigurationProps> = ({ cluster })
         values,
         validateField,
       }: FormikProps<ClusterConfigurationValues>) => {
-        const handleSubmit = (e: React.MouseEvent) => {
-          setFieldValue('submitType', (e.target as HTMLButtonElement).name);
+        const handleSubmit = async (e: React.MouseEvent) => {
```
mareklibra, line 233: Can we use different name? To avoid conflict with the enclosing context.

src/components/clusters/newClusterModal.tsx
```diff
@@ -73,18 +57,30 @@ export const NewClusterModal: React.FC<NewClusterModalProps> = ({ closeModal })
...
   );
 
   const handleSubmit = async (
     values: ClusterCreateParams,
     formikActions: FormikHelpers<ClusterCreateParams>,
   ) => {
     formikActions.setStatus({ error: null });
+
+    // async validation for cluster name - run only on submit
+    try {
```
mareklibra, line 74: What about running it twice? Get list of clusters when opening the page, keep validating against it and then (if really needed) refresh when submitting.

src/components/clusterConfiguration/ClusterConfiguration.tsx
```diff
@@ -112,10 +131,14 @@ const findMatchingSubnet = (
   return matchingSubnet || 'No subnets available';
 };
 
+type ClusterConfigurationProps = {
+  cluster: Cluster;
+};
+
 const ClusterConfiguration: React.FC<ClusterConfigurationProps> = ({ cluster }) => {
+  const [submitType, setSubmitType] = React.useState('save');
```
rawagner, line 139: this looks weird to me. Why not just pass the submit type to `handleSubmit` func instead ?

src/components/clusters/newClusterModal.tsx
jtomasek: In a separate change we could add a field level validation for the name field to do the async validation onBlur.

src/components/clusterConfiguration/ClusterConfiguration.tsx
jtomasek: submitType also drives the variant of validationSchema prop, so I had to push the state for this to the component level.

src/components/clusterConfiguration/ClusterConfiguration.tsx
jtomasek: I'd have to move the whole form into a subcomponent. I've added a TODO note for that.

mareklibra: ok, I believe this will not be a performance issue for M1

PR #154, opened 2020-06-01
Allow rootless podman build

mareklibra:
Intentionally keeping the `1.17` version (latest is `1.18`), will be upgraded later to mitigate risks for now.


mareklibra:
Merged into #165

PR #158, opened 2020-06-01
Advanced networking

jtomasek:
Depends on https://github.com/openshift-metal3/facet/pull/153

### DIFF
--- a/src/components/clusterConfiguration/ClusterConfiguration.tsx
+++ b/src/components/clusterConfiguration/ClusterConfiguration.tsx
@@ -272,25 +272,6 @@ const ClusterConfiguration: React.FC<ClusterConfigurationProps> = ({ cluster }) 
                       <TextContent>
                         <Text component="h2">Networking</Text>
                       </TextContent>
-                      <InputField
-                        name="clusterNetworkCidr"
-                        label="Cluster Network CIDR"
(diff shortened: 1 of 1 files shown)


src/components/clusterConfiguration/ClusterConfiguration.tsx
```diff
@@ -327,6 +308,28 @@ const ClusterConfiguration: React.FC<ClusterConfigurationProps> = ({ cluster })
...
                         validate={(value: string) => validateVIP(subnetHostMap, values, value)}
                       />
+                      <TextContent>
+                        <Text component="h2">Advanced Networking</Text>
+                      </TextContent>
+                      <InputField
+                        name="clusterNetworkCidr"
+                        label="Cluster Network CIDR"
+                        helperText="IP address block from which Pod IPs are allocated This block must not overlap with existing physical networks. These IP addresses are used for the Pod network, and if you need to access the Pods from an external network, configure load balancers and routers to manage the traffic."
+                        isRequired
```
rawagner, line 318: advanced fields are not required, right ?

jtomasek: I believe they are required to be filled in, there will be defaults provided on cluster creation but if user removes that value the deployment won't proceed IIUC. @filanov ?

andybraren: I just got access to the creation form for OpenShift Dedicated clusters in OCM and noticed that they make these "advanced" fields optional but with placeholder text to show the default value that will be used if the user doesn't add anything else. What do you think of using the same pattern?

image

mareklibra:
#153 is merged now, rebase needed

PR #163, opened 2020-06-02
Add statusUpdatedAt to cluster and host status popover

jtomasek:


src/components/hosts/HostStatus.tsx
```diff
@@ -80,11 +82,20 @@ const HostStatus: React.FC<HostStatusProps> = ({ host }) => {
     ),
     [status, progressInfo, statusInfo],
   );
+  const footerContent = React.useMemo(
+    () => (
+      <Tooltip content={getHumanizedDateTime(statusUpdatedAt)} position={TooltipPosition.right}>
+        <small>Updated {hdate.relativeTime(statusUpdatedAt || '')}</small>
```
mareklibra, line 88: likewise

andybraren:
Is the output of `getHumanizedDateTime` similar to the timestamps seen in the Events page of the OCP Web Console? On that page it looks like after `10 minutes ago` the format becomes `Jun 2, 5:30pm`, which may be good to align with.

jtomasek:
We're moving from relative time to actual time with this change:
Screenshot 2020-06-03 at 10 09 14

src/components/clusters/ClusterStatus.tsx
```diff
@@ -25,18 +26,28 @@ const getStatusIcon = (status: Cluster['status']) => {
...
-  statusInfo?: string;
+  cluster: Cluster;
 };
 
-const ClusterStatus: React.FC<ClusterStatusProps> = ({ status, statusInfo }) => {
+const ClusterStatus: React.FC<ClusterStatusProps> = ({ cluster }) => {
+  const { status, statusInfo, statusUpdatedAt } = cluster;
   const title = CLUSTER_STATUS_LABELS[status];
   const icon = getStatusIcon(status);
   const headerContent = React.useMemo(() => <div>{title}</div>, [title]);
```
rawagner, line 36: IMO all these useMemos are not needed and i'd say its actually _harmful_ to the app performance. These are simple values which do not need to be memoized.

jtomasek: done

PR #168, opened 2020-06-02
Validate IP Address against regex, move all ip validations back to schema

rawagner:


src/components/clusterConfiguration/ClusterConfiguration.tsx
```diff
@@ -97,6 +98,13 @@ const validateVIP = (
   values: ClusterConfigurationValues,
   value: string,
 ) => {
+  if (!value) {
```
mareklibra, line 101: Is there a reason why we don't use validation schemas here?

src/components/ui/formik/validationSchemas.ts
```diff
@@ -3,7 +3,7 @@ import * as Yup from 'yup';
...
+export const IP_ADDRESS_REGEX = /^(((([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5]))|([0-9a-f]{1,4}:){7,7}[0-9a-f]{1,4}|([0-9a-f]{1,4}:){1,7}:|([0-9a-f]{1,4}:){1,6}:[0-9a-f]{1,4}|([0-9a-f]{1,4}:){1,5}(:[0-9a-f]{1,4}){1,2}|([0-9a-f]{1,4}:){1,4}(:[0-9a-f]{1,4}){1,3}|([0-9a-f]{1,4}:){1,3}(:[0-9a-f]{1,4}){1,4}|([0-9a-f]{1,4}:){1,2}(:[0-9a-f]{1,4}){1,5}|[0-9a-f]{1,4}:((:[0-9a-f]{1,4}){1,6})|:((:[0-9a-f]{1,4}){1,7}|:)|fe80:(:[0-9a-f]{0,4}){0,4}%[0-9a-z]{1,}|::(ffff(:0{1,4}){0,1}:){0,1}((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])|([0-9a-fA-F]{1,4}:){1,4}:((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]))$/;
```
mareklibra, line 6: Already wrapped by `ipValidationSchema()`, if needed

rawagner:
@mareklibra regex way was just a quick fix for @jtomasek . I've updated the PR using the Yup validation, ptal

src/components/ui/formik/validationSchemas.ts
andybraren: `// Future bug-fixer: Beer on me! (mlibra)`
😂

src/components/clusterConfiguration/ClusterConfiguration.tsx
```diff
@@ -217,7 +204,11 @@ const ClusterConfiguration: React.FC<ClusterConfigurationProps> = ({ cluster })
   return (
     <Formik
       initialValues={initialValues}
-      validationSchema={submitType === 'install' ? installValidationSchema : validationSchema}
+      validationSchema={
```
mareklibra, line 207: I think this should be memoized

PR #169, opened 2020-06-02
Change event polling interval

mareklibra:


src/config/constants.ts
```diff
@@ -14,7 +14,8 @@ export const OPENSHIFT_VERSION_OPTIONS: OpenshiftVersionOptionType[] = [
 export const CLUSTER_MANAGER_SITE_LINK =
   'https://cloud.redhat.com/openshift/install/metal/user-provisioned';
 
-export const POLLING_INTERVAL = 10000;
+export const POLLING_INTERVAL = 10 * 1000;
```
mareklibra, line 17: A cluster polling interval should be increased too. What about 60 seconds?

src/config/constants.ts
```diff
@@ -14,7 +14,8 @@ export const OPENSHIFT_VERSION_OPTIONS: OpenshiftVersionOptionType[] = [
...
+export const EVENTS_POLLING_INTERVAL = 60 * 1000;
```
mareklibra, line 18: I am not sure about such long interval. I think 30 seconds would be reasonable delay. @jtomasek , wdyt?

jtomasek: IMO the cluster polling interval should be kept at 10s, we need to keep the UI up to date - state changes etc.
60s for events is probably fine

mareklibra: Ok, changing the cluster to 10 secs to keep the rendered info up-to-date.

I still think we should decrease the time for event polling at least once we move the events to a modal.

PR #170, opened 2020-06-03
Temporarily remove sidebar

mareklibra:
Sidebar will be back in the future.


src/components/App.tsx
```diff
@@ -19,7 +19,6 @@ const App: React.FC = () => (
       <BackgroundImage />
       <Page
         header={<Header />}
-        sidebar={<Sidebar />}
```
jtomasek, line 22: Nit: Could be just commented out

mareklibra: done

PR #171, opened 2020-06-03
Remove ID, add Base DNS Domain to ClusterTable

mareklibra:


src/selectors/clusters.tsx
```diff
@@ -36,7 +36,7 @@ const clusterToClusterTableRow = (cluster: Cluster): IRow => {
         ),
         sortableValue: name,
       } as HumanizedSortable,
-      id,
+      baseDnsDomain,
```
jtomasek, line 39:
```suggestion
      baseDnsDomain || DASH,
```

mareklibra: done

PR #173, opened 2020-06-03
Pull secret is not revealed once set

mareklibra:
---

### DIFF
--- a/src/components/clusterConfiguration/ClusterConfiguration.tsx
+++ b/src/components/clusterConfiguration/ClusterConfiguration.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { Formik, FormikProps, FormikHelpers } from 'formik';
 import * as Yup from 'yup';
 import { Link } from 'react-router-dom';
+import _ from 'lodash';
 import {
   Form,
   PageSectionVariants,
@@ -21,7 +22,7 @@ import { Netmask } from 'netmask';
 import ClusterToolbar from '../clusters/ClusterToolbar';
 import PageSection from '../ui/PageSection';
 import { ToolbarButton, ToolbarText } from '../ui/Toolbar';
-import { InputField, TextAreaField } from '../ui/formik';
+import { InputField, TextAreaField, TextAreaSecretField } from '../ui/formik';
 import GridGap from '../ui/GridGap';
 import { Cluster, ClusterUpdateParams, Inventory } from '../../api/types';
 import { patchCluster, postInstallCluster, getClusters } from '../../api/clusters';
@@ -121,6 +122,7 @@ const ClusterConfiguration: React.FC<ClusterConfigurationProps> = ({ cluster }) 
   const [submitType, setSubmitType] = React.useState('save');
   const dispatch = useDispatch();
   const [alerts, dispatchAlertsAction] = React.useReducer(alertsReducer, []);
+  const [isPullSecretEdit, setPullSecretEdit] = React.useState(!cluster.pullSecretSet);
 
   const hostnameMap: { [id: string]: string } =
     cluster.hosts?.reduce((acc, host) => {
@@ -150,7 +152,7 @@ const ClusterConfiguration: React.FC<ClusterConfigurationProps> = ({ cluster }) 
     serviceNetworkCidr: cluster.serviceNetworkCidr || '',
     apiVip: cluster.apiVip || '',
     ingressVip: cluster.ingressVip || '',
-    pullSecret: cluster.pullSecret || '',
+    pullSecret: '',
     sshPublicKey: cluster.sshPublicKey || '',
     hostSubnet: findMatchingSubnet(cluster.ingressVip, cluster.apiVip, hostSubnets),
   };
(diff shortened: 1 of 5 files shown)


src/components/clusterConfiguration/ClusterConfiguration.tsx
```diff
@@ -173,6 +172,10 @@ const ClusterConfiguration: React.FC<ClusterConfigurationProps> = ({ cluster })
 
     // update the cluster validation
     try {
+      if (!values.pullSecret) {
```
jtomasek, line 175: This will get removed once the pull secret is not available from API, right? If so, a TODO would be nice

src/components/ui/formik/TextAreaField.tsx
```diff
@@ -4,11 +4,19 @@ import { FormGroup, TextArea } from '@patternfly/react-core';
...
+  helperText,
+  isRequired,
+  children,
+  ...props
+}) => {
   const [field, { touched, error }] = useField(props.name);
   const fieldId = getFieldId(props.name, 'input');
   const isValid = !(touched && error);
   const errorMessage = !isValid ? error : '';
+  const handleAreaChange = React.useCallback((value, event) => field.onChange(event), [field]);
```
jtomasek, line 18: I think using useCallback here is not helping the performance
https://kentcdodds.com/blog/usememo-and-usecallback

src/components/ui/formik/TextAreaSecretField.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const [isUpdate, setUpdate] = React.useState(false);
+  const fieldId = getFieldId(name, 'input');
+  const handleClick = React.useCallback(() => {
+    setUpdate(!isUpdate);
+    onReset && onReset();
+  }, [setUpdate, isUpdate, onReset]);
+
+  if (isSet && !isUpdate) {
+    return (
+      <FormGroup fieldId={fieldId} label={label} helperText={helperTextHidden}>
```
jtomasek, line 23:
```suggestion
      <FormGroup fieldId={fieldId} label={label} helperText={helperTextHidden} isRequired={props.isRequired}>
```

src/components/clusterConfiguration/ClusterConfiguration.tsx
mareklibra: This will remain here since the `pullSecret` will remain among `cluster-update-params`. With this patch, it is not read from `Cluster` anymore, but remains among formik's fields since to support the "update" flow. The `delete` here ensures that the backend will not update its value when not intended to.

src/components/ui/formik/TextAreaSecretField.tsx
mareklibra: This has probably no practical effect, but ok, added

src/components/clusterConfiguration/ClusterConfiguration.tsx
```diff
@@ -173,6 +172,10 @@ const ClusterConfiguration: React.FC<ClusterConfigurationProps> = ({ cluster })
...
+        // avoid update by backend
+        delete values.pullSecret;
+      }
       const { data } = await patchCluster(cluster.id, values);
       dispatch(updateCluster(data));
```
jtomasek, line 180: setProvidingPullSecret(false);

mareklibra: > I think we should hide the text area every time the form is successfully submitted.

And it already works like that. The user can see pull secret until the form is submitted. Once the cluster is updated (Save draft), its new configuration is returned, newly including `providing_pull_secret: true` which is subsequently passed down to the component which tuns into "hiding" mode - the secret is not shown. 

When clicking on "Change", the secret is not rendered but the user is able to either provide new value or leave it.

mareklibra: > Related to comment above, I'd prefer keeping the values below intact and doing this instead:

Yeah, it makes sense to create "a modified copy" right before patching. Will do that.

mareklibra: > By doing this we also get ability to reset pull secret field (submit it with '' value) by changing this (line 175) condition to

I think it's is unnecessary logic. Recent use-case can be handled without introducing yet another boolean to drive the flow. In addition, the `pullSecret` is a required param so we should either keep it untouched or pass valid value. If this ever changes, we can tweak the logic round "removing" the field from the cluster-update-params.

src/components/clusterConfiguration/ClusterConfiguration.tsx
```diff
@@ -173,8 +174,14 @@ const ClusterConfiguration: React.FC<ClusterConfigurationProps> = ({ cluster })
 
     // update the cluster validation
     try {
-      const { data } = await patchCluster(cluster.id, values);
+      const params = { ...values };
+      if (!params.pullSecret) {
+        // avoid update to empty value by backend
+        delete params.pullSecret;
+      }
```
jtomasek, line 181: Even if we reset the field value when hiding the textarea, I'd prefer setting the pull secret only when `isPullSecretEdit` is enabled regardless of the actual value of the pullSecret. This would eventually allow reseting the value with empty string.
```suggestion
      const params = isPullSecretEdit ? _.omit(values, ['pullSecret']) : values;
```

mareklibra: vice versa, but ok

PR #178, opened 2020-06-05
Move cluster events into modal

jtomasek:
Modal is opened by 'View Cluster Event Hostory' in the cluster toolbar

### DIFF
--- a/src/components/clusters/clusterEventsModal.tsx
+++ b/src/components/clusters/clusterEventsModal.tsx
@@ -1,0 +1,48 @@
+import React from 'react';
+import { Button, Modal } from '@patternfly/react-core';
+import { ToolbarButton } from '../ui/Toolbar';
+import EventListFetch from '../fetching/EventListFetch';
+import { Cluster } from '../../api/types';
+
+type ClusterEventsModalButtonProps = React.ComponentProps<typeof Button> & {
+  ButtonComponent?: typeof Button | typeof ToolbarButton;
+  onClick?: () => void;
+  clusterId: Cluster['id'];
+};
+
+export const ClusterEventsModalButton: React.FC<ClusterEventsModalButtonProps> = ({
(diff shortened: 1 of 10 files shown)


src/components/clusters/clusterEventsModal.tsx
```diff
@@ -0,0 +1,48 @@
...
+};
+
+export const ClusterEventsModal: React.FC<ClusterEventsModalProps> = ({
+  clusterId,
+  onClose,
+  isOpen,
+}) => {
+  return (
+    <Modal title="Cluster Events" isOpen={isOpen} onClose={onClose} isFooterLeftAligned isLarge>
+      <EventListFetch entityId={clusterId} entityKind="cluster" />
```
mareklibra, line 45: As an user, I miss the Close button (primary) in footer. The cross in top-right corner is not enough for me :-)

mareklibra, line 45: The table in modal is rendered always with vertical scrollbar, no matter the content can fit.

jtomasek: Yeah, this seems like a bug in modal/table. I'd like to try to fix it as a separate issue. It is most likely PF problem

jtomasek: I'll fix it as part of https://github.com/openshift-metal3/facet/pull/179 to avoid rebase issues as I am consolidating the modal there

jtomasek: Fixed as part of https://github.com/openshift-metal3/facet/pull/179/commits

mareklibra: Agree, it's PF issue. But we can workaround it in our code base, it can be fixed by playing with `height`.

PR #179, opened 2020-06-05
Host events modal

jtomasek:
Move host events into modal triggered by host kebab action
Unify hosts and cluster events modal

Depends on #178


src/components/hosts/HostsTable.tsx
```diff
@@ -196,6 +198,14 @@ const HostsTable: React.FC<HostsTableProps> = ({ cluster }) => {
     [cluster.id, dispatch, addAlert],
   );
 
+  const onViewHostEvents = React.useCallback(
```
mareklibra, line 201: nit: the `useCallback` has no practical effect in the case of this component

PR #184, opened 2020-06-08
Add axios interceptor to unify reason/message under a message key

jtomasek:


src/components/clusterConfiguration/ClusterConfiguration.tsx
```diff
@@ -178,7 +178,7 @@ const ClusterConfiguration: React.FC<ClusterConfigurationProps> = ({ cluster })
     } catch (e) {
       handleApiError<ClusterUpdateParams>(e, () =>
         dispatchAlertsAction(
-          addAlert({ title: 'Failed to update the cluster', message: e.response?.data?.reason }),
+          addAlert({ title: 'Failed to update the cluster', message: e.response?.data?.message }),
```
mareklibra, line 181: I wonder when this will be changed.
Would it make sense to wrap `e.response?.data?.message` into a selector and chain retrieval from various cases there? Sort of:
```
const getErrorMessage = (e) => e.response?.data?.message ||  e.response?.data?.reason;
```

We can simply add more variants when once they appear.

jtomasek: I agree that selector solution is probably better than axios interceptor as it makes debugging harder as the actual data is different than what we see in the response in browser dev tools. I'll switch to using selector

jtomasek: Done

PR #180, opened 2020-06-05
Instruct user to set DNS

mareklibra:
New modal dialog is open when the user hits "Launch OpenShift Console" button.

New "(Link not working)" is added close to the console link.

Work in progress, TODO:
- [x] text
- [x] placement
- [x] external IP

PR #191, opened 2020-06-09
Allow copy&paste of the /etc/hosts hint

mareklibra:
Follow-up for #180

### DIFF
--- a/src/components/clusterDetail/ClusterCredentials.tsx
+++ b/src/components/clusterDetail/ClusterCredentials.tsx
@@ -7,7 +7,7 @@ import {
   clipboardCopyFunc,
   Popover,
 } from '@patternfly/react-core';
-import { ExternalLinkSquareAltIcon, InfoCircleIcon } from '@patternfly/react-icons';
+import { InfoCircleIcon, ExternalLinkAltIcon } from '@patternfly/react-icons';
 
 import { LoadingState, ErrorState } from '../ui/uiState';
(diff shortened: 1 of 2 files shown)


mareklibra:
---

mareklibra:
@andybraren , what about aligning the icons for these two links to the same side?

mareklibra:
I have added more records to the list

jtomasek:
The text in the popover overflows. We'll need to set overflow to `<pre>` or how about just ditching the popover and open the ConsoleModal? IMHO it would make the experience a bit more unified. WDYT?

Screenshot 2020-06-09 at 15 19 22

andybraren:
Yep, that's a lot of content, time for a modal. 🙂

> @andybraren , what about aligning the icons for these two links to the same side?

I agree it looks a little odd, but external link icons on the right and info icons on the left is the PatternFly [pattern](https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/embedded-assistance).

If we change the external link icon to be `fa-external-link-alt` available in the (brand new!) [Icons page](https://www.patternfly.org/v4/design-guidelines/styles/icons#all-icons) it should look less heavy:

image

mareklibra:
I moved the hint into modal and replaced `<pre>` by simple `<div>` for the content to allow reasonable wrapping on smaller displays. Copy&paste still works nice (means: with the correct line separators).

PR #194, opened 2020-06-09
ClusterConfiguration validation: Ingress and API VIPs must differ

mareklibra:


src/components/ui/formik/validationSchemas.ts
```diff
@@ -56,6 +59,23 @@ export const vipValidationSchema = (hostSubnets: HostSubnets, values: ClusterCon
     return !!subnet?.contains(value) && value !== subnet?.broadcast && value !== subnet?.base;
   });
 
+const vipUniqueValidatinoSchema = (hostSubnets: HostSubnets, values: ClusterConfigurationValues) =>
```
jtomasek, line 62: Typo
```suggestion
const vipUniqueValidationSchema = (hostSubnets: HostSubnets, values: ClusterConfigurationValues) =>
```

src/components/ui/formik/validationSchemas.ts
```diff
@@ -56,6 +59,23 @@ export const vipValidationSchema = (hostSubnets: HostSubnets, values: ClusterCon
...
+  Yup.string().test(
```
jtomasek, line 63: This could be replaced with
```
yup.string().notOneOf([yup.ref('apiVip')], "API VIP and Ingress VIP must be different.")
```

mareklibra: I want reuse one schema for both fields. And skip empty value. Isn't recent implementation simpler?

src/components/ui/formik/validationSchemas.ts
mareklibra: done

src/components/ui/formik/validationSchemas.ts
jtomasek: OK, both works. I just fell like we're diving into creating generic functions for validation while we have only single form which for which we could tailor a simple validation schema :)

PR #195, opened 2020-06-09
Add Red Hat favicon

jtomasek:


mareklibra:
This might be the reason for OKD to create a fork in the future.

PR #197, opened 2020-06-09
Fix ClusterToolbar

jtomasek:


src/components/ui/Toolbar/ToolbarSecondaryGroup.tsx
```diff
@@ -1,11 +1,8 @@
 import React from 'react';
-import { ToolbarItem, ToolbarItemVariant, ToolbarGroup } from '@patternfly/react-core';
+import { ToolbarGroup } from '@patternfly/react-core';
 
 const ToolbarSecondaryGroup: React.FC = ({ children }) => (
-  <>
-    <ToolbarItem variant={ToolbarItemVariant.separator} visiblity={{ md: 'hidden' }} />
-    <ToolbarGroup alignment={{ md: 'alignRight' }}>{children}</ToolbarGroup>
-  </>
+  <ToolbarGroup alignment={{ md: 'alignRight' }}>{children}</ToolbarGroup>
```
mareklibra, line 5: Do we even need this simple `ToolbarSecondaryGroup` wrapper?

jtomasek: I'd say it still brings some value considering it is used several times.

mareklibra: I will not be against it :-)

PR #198, opened 2020-06-09
Disable cluster delete

mareklibra:
Depends on 
- [x] #193


src/selectors/clusters.tsx
```diff
@@ -55,6 +55,9 @@ const clusterToClusterTableRow = (cluster: Cluster): IRow => {
   };
 };
 
+export const getClusterTableStatusCell = (rowData: IRowData) =>
+  rowData?.cells?.[3] as HumanizedSortable;
```
mareklibra, line 59: We should start to be careful when referencing cells. The table structure seems to be not stable as expected to be at the beginning.

PR #203, opened 2020-06-10
Pull Secret is not required for installation

mareklibra:
If already set on backend.

Otherwise it's still required.


src/components/ui/formik/validationSchemas.ts
```diff
@@ -37,6 +37,13 @@ export const validJSONSchema = Yup.string().test(
   },
 );
 
+export const pullSecretKnownOrRequired = (values: ClusterConfigurationValues) =>
+  Yup.string().test(
+    'pull-secret-known-or-required',
+    'Pull secret ust be provided.',
```
jtomasek, line 43: Typo
```suggestion
    'Pull secret must be provided.',
```
Or to match the other fields message: 'Required to install the cluster.'

src/components/clusterConfiguration/ClusterConfiguration.tsx
```diff
@@ -154,6 +155,7 @@ const ClusterConfiguration: React.FC<ClusterConfigurationProps> = ({ cluster })
     ingressVip: cluster.ingressVip || '',
     pullSecret: '',
     sshPublicKey: cluster.sshPublicKey || '',
+    pullSecretSet: !!cluster.pullSecretSet, // private (read-only), for validations only, set by backend
```
jtomasek, line 158: Since we're introducing this, we don't need to keep the value in the component in the state (isPullSecretEdit, setPullSecretEdit) any more. We can replace those with `values.pullSecretSet` and `setFieldValue('pullSecretSet',...)

mareklibra: done

PR #207, opened 2020-06-10
Add mechanism to identify cluster is ready for installation

jtomasek:
- Add ClusterValidationSection to ClusterToolbar to expose current
  cluster configuration problems
- Add cluster validations which are calculated on every cluster update
- Simplify install validation mechanism by moving the logic from the
  form into the cluster based validations
- Separate the 'Save Changes' and 'Install Cluster' logic
- Add 'Discard Changes' button to allow rolling back form changes
- Refactor utility functions from ClusterConfiguration into a separate
  module

Screenshot 2020-06-10 at 23 43 26
Screenshot 2020-06-10 at 23 44 06
Screenshot 2020-06-10 at 23 44 33

### DIFF
--- a/src/components/clusterConfiguration/ClusterConfiguration.tsx
+++ b/src/components/clusterConfiguration/ClusterConfiguration.tsx
@@ -13,11 +13,12 @@ import {
   GridItem,
   TextVariants,
   Spinner,
+  Button,
 } from '@patternfly/react-core';
-import { ExternalLinkAltIcon, ExclamationCircleIcon } from '@patternfly/react-icons';
-import { global_danger_color_100 as dangerColor } from '@patternfly/react-tokens';
+import { ExternalLinkAltIcon, WarningTriangleIcon, CheckCircleIcon } from '@patternfly/react-icons';
+import { global_success_color_100 as successColor } from '@patternfly/react-tokens';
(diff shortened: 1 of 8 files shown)


src/components/clusterConfiguration/ClusterConfiguration.tsx
```diff
@@ -269,7 +189,7 @@ const ClusterConfiguration: React.FC<ClusterConfigurationProps> = ({ cluster })
                       </TextContent>
                       <InputField label="Cluster Name" name="name" isRequired />
                       <InputField
-                        label="Base DNS domain"
+                        label="Base DNS Domain"
```
mareklibra, line 192: I think we are mixing these capitalized and non capitalized labels throughout the screens. We should pass through all strings and unify them

src/config/constants.ts
```diff
@@ -22,7 +22,7 @@ export const EVENTS_POLLING_INTERVAL = 10 * 1000;
 
 export const HOST_ROLES = ['worker', 'master'];
 // Without undefined. Otherwise must conform generated Host['roles'] - see api/types.ts
-export type HostRolesType = 'master' | 'worker';
+export type Role = 'master' | 'worker' | undefined;
```
mareklibra, line 25: There was reason for introducing this type without `undefined`. Otherwise let's use directly `Host['roles']`

PR #210, opened 2020-06-11
Add error handling for validateYupSchema

jtomasek:
And use TextList component in ClusterValidationSection

### DIFF
--- a/src/components/clusterConfiguration/clusterValidations.ts
+++ b/src/components/clusterConfiguration/clusterValidations.ts
@@ -49,9 +49,18 @@ export const validateRequiredFields = (cluster: Cluster) => {
   try {
     validateYupSchema<ClusterConfigurationValues>(values, installValidationSchema, true);
   } catch (err) {
-    const errors = yupToFormErrors(err);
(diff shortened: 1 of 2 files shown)


src/components/clusterConfiguration/clusterValidations.ts
```diff
@@ -49,9 +49,18 @@ export const validateRequiredFields = (cluster: Cluster) => {
   try {
     validateYupSchema<ClusterConfigurationValues>(values, installValidationSchema, true);
   } catch (err) {
-    const errors = yupToFormErrors(err);
-    const errorFields = Object.keys(errors).map((field: string) => CLUSTER_FIELD_LABELS[field]);
-    return `Not all required cluster properties are configured yet: ${errorFields.join(', ')}.`;
+    if (err.name === 'ValidationError') {
```
rawagner, line 52: instanceof ValidationError would be a safer approach as we wont have to use string
```suggestion
    if (err instanceof ValidationError) {
```

jtomasek: Done

PR #218, opened 2020-06-15
Introduce facet-lib

mareklibra:
Our code is about to be reused by other applications.


jtomasek:
- There seems to be no clear way to develop on the library
- I think the library should not re-export all modules all the way to the top level as it causes name collisions and cyclic dependencies
- Styles and images are currently broken

mareklibra:
> There seems to be no clear way to develop on the library

We will need modify both the `facet-lib` for auto-rebuild and the `facet` to accept changes in node_modules.

We can leverage npm linking or simply `ln -s` from facet to built `dist`.

> I think the library should not re-export all modules all the way to the top level as it causes name collisions and cyclic dependencies

We should avoid cyclic dependencies inside the library. If any, we should refactor the code.

I was thinking about "namespaces", like wraping exports inside i.e. `api` object which will be exported from the library out.

> Styles and images are currently broken

Which images?

Styles need to be fixed, it's about webpack loader. Maybe we can start using styled components. @rawagner , what do you think about them in context of integration with `cloud.*` application?

jtomasek:
I think we probably should revert this until we have addressed these^ as otherwise the development is blocked.

mareklibra:
> I think we probably should revert this until we have addressed these^ as otherwise the development is blocked.

I don't agree. We are just looking for more comfortable solution than
```
cd [facet-lib] 
yarn build ; rm -rf ../facet/node_modules/facet-lib/dist ; cp -r dist ../facet/node_modules/facet-lib/
``

PR #219, opened 2020-06-15
Expose UI version

jtomasek:
Use latest git tag and git sha exposed to the UI on build time via environment
variables. Display the UI version in the masthead.

### DIFF
--- a/.github/workflows/buildAndPush.yaml
+++ b/.github/workflows/buildAndPush.yaml
@@ -19,10 +19,15 @@ jobs:
         uses: borales/actions-yarn@v2.0.0
         with:
           cmd: lint
+      - name: Get tag
+        id: get_tag
+        run: echo ::set-env name=GIT_TAG::${GITHUB_REF/refs\/tags\//}
       - name: Build code
         uses: borales/actions-yarn@v2.0.0
         env:
           NODE_OPTIONS: '--max-old-space-size=8192'
+          REACT_APP_GIT_SHA: ${{ github.sha }}
+          REACT_APP_VERSION: ${{ env.GIT_TAG }}
         with:
           cmd: build
       - name: Publish to quay.io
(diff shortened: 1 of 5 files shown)


src/components/ui/Header.tsx
```diff
@@ -37,6 +37,7 @@ const Header: React.FC = () => {
           >
             Provide feedback <ExternalLinkAltIcon />
           </Button>
+          {process.env.REACT_APP_VERSION}
```
mareklibra, line 40: Let's wrap this into an utility function

.github/workflows/buildAndPush.yaml
```diff
@@ -19,10 +19,15 @@ jobs:
         uses: borales/actions-yarn@v2.0.0
         with:
           cmd: lint
+      - name: Get release version
+        id: get_version
+        run: echo ::set-env name=REACT_APP_VERSION::$(git describe --abbrev=0)
```
mareklibra, line 24: This effectively takes latest `git tag`, I think more relevant is referencing via git's `sha`

src/components/ui/Header.tsx
mareklibra: Is it available at runtime?

mareklibra:
Per discussion, the tags when:
- merged to master: `laterst`, `sha`
- tagged in a stable branch: `tag`, `sha` (which has probably no use)

.github/workflows/buildAndPush.yaml
```diff
@@ -33,7 +38,7 @@ jobs:
           password: ${{ secrets.QUAYIO_OCPMETAL_PASSWORD }}
           registry: quay.io
           dockerfile: Dockerfile
-          tags: 'latest,${{ github.sha }}'
+          tags: 'latest,${{ github.sha }},${{ env.GIT_TAG }}'
```
mareklibra, line 41: we should not use this here but in "on tag" only. Makes probably no sense here, just confusing.

.github/workflows/buildAndPush.yaml
```diff
@@ -42,4 +47,4 @@ jobs:
           password: ${{ secrets.QUAYIO_OCPMETAL_PASSWORD }}
           registry: quay.io
           dockerfile: Dockerfile.cypress
-          tags: 'latest,${{ github.sha }}'
+          tags: 'latest,${{ github.sha }},${{ env.GIT_TAG }}'
```
mareklibra, line 50: likewise

jtomasek: done

.github/workflows/buildAndPush.yaml
jtomasek: done

PR #221, opened 2020-06-16
MGMT-1166 Added release.yaml to github workflow

eranco74:
This workflow will build the repository container images upon git tag creation and publish it to ocpmetal registry with the matching tag

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,48 @@
+name: Build and push
+
+on:
+  push:
+    tags:
+      - "*"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v1
+      - name: Install dependencies
+        uses: borales/actions-yarn@v2.0.0
+        with:
+          cmd: install
+      - name: Call linter
+        uses: borales/actions-yarn@v2.0.0
+        with:
+          cmd: lint
+      - name: Build code
+        uses: borales/actions-yarn@v2.0.0
+        env:
+          NODE_OPTIONS: '--max-old-space-size=8192'
+        with:
+          cmd: build
+      - name: Get tag
+        id: get_tag
+        run: echo ::set-env name=GIT_TAG::${GITHUB_REF/refs\/tags\//}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,48 @@
+name: Build and push
+
+on:
+  push:
```
jtomasek, line 4: I guess once the stable branch is created we'll want to add it here, right?
```
branches:
  - stable
```

.github/workflows/release.yaml
```diff
@@ -0,0 +1,48 @@
...
+          cmd: lint
+      - name: Build code
+        uses: borales/actions-yarn@v2.0.0
+        env:
+          NODE_OPTIONS: '--max-old-space-size=8192'
+        with:
+          cmd: build
+      - name: Get tag
+        id: get_tag
+        run: echo ::set-env name=GIT_TAG::${GITHUB_REF/refs\/tags\//}
```
jtomasek, line 30: How is GITHUB_REF populated?

jtomasek: I see now... https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#default-environment-variables

mareklibra: Provided by GitHub: https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables

.github/workflows/release.yaml
eranco74: Unsure, the plan is to create a tag for each release.
and deploy using that tag.
No real usage for the "stable" image.

.github/workflows/release.yaml
eranco74: Uou can see it in action here (created a tag just for testing):
https://github.com/eranco74/assisted-installer/actions/runs/136882098

.github/workflows/release.yaml
mareklibra: IMO, we can keep it per-every-tag. We will just not tag master.

jtomasek: ok, sounds good

PR #235, opened 2020-06-25
Fix service labels and use DetailList in AboutModal

jtomasek:
Depends on https://github.com/mareklibra/facet-lib/pull/46
Screenshot 2020-06-25 at 10 56 46

### DIFF
--- a/src/components/AboutModal.tsx
+++ b/src/components/AboutModal.tsx
@@ -3,13 +3,11 @@ import {
   Button,
   AboutModal as PFAboutModal,
   TextContent,
-  TextList,
-  TextListItem,
   ButtonVariant,
 } from '@patternfly/react-core';
 import { GIT_SHA, VERSION, SERVICE_LABELS } from '../config/standalone';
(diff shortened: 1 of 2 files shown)


src/components/AboutModal.tsx
```diff
@@ -58,18 +57,21 @@ const AboutModal: React.FC<AboutModalProps> = ({ isOpen, onClose }) => {
...
-          </TextListItem>
-          {Object.keys(versions).map((version) => (
-            <>
-              <TextListItem component="dt">{SERVICE_LABELS[version] || version}</TextListItem>
-              <TextListItem component="dd">{versions[version]}</TextListItem>
-            </>
-          ))}
-        </TextList>
+        <DetailList>
+          <>
```
mareklibra, line 61: Is this needed?

jtomasek: yes because DetailList expect children to be either ReactChild or ReactChild[]
`children: ReactChild | (ReactChild | undefined)[];`

mareklibra: The resulting type would be `{ children: (Element | Element[])[]; }`. I agree that this is the easiest workaround. Or wrapping just the generated array within `<>`

PR #236, opened 2020-06-25
Update E2E tests after M1

mareklibra:
Depends on: 
- [x] https://github.com/mareklibra/facet-lib/pull/48

There is a bunch of `TODOs` with ideas identified when writing the tests. They can be addressed in follow-ups. With these tests, we have quite good coverage.

As a next higher-priority task, we need to figure out passing of a valid pull-secret so we can actually start the cluster installation.


mareklibra:
tests

PR #254, opened 2020-07-21
QE E2E

ukalifon:
QE E2E 

These are the 2 scripts that are run by the Ansible automation:
1) The 1st script handles the creation of the cluster in the UI up until the 
   downloading of the ISO.
2) The 2nd script runs after the ISO was downloaded and the hosts booted

To prevent it from accidentally being run in other CIs, it is located in
a non-default directory. To run it: 

export CYPRESS_BASE_URL=https://...
export CYPRESS_CLUSTER_NAME=cypress
export CYPRESS_DNS_DOMAIN_NAME=qe.lab.redhat.com
export CYPRESS_API_VIP=192.168.123.5
export CYPRESS_INGRESS_VIP=192.168.123.10
export CYPRESS_SSH_PUB_KEY=$(cat ~/.ssh/id_rsa.pub)
export CYPRESS_PULL_SECRET='...'
export CYPRESS_NUM_MASTERS=3
export CYPRESS_NUM_WORKERS=2
mv cypress/integration-qe/e2e-createNewCluster.spec.js cypress/integration/
mv cypress/integration-qe/e2e-enterClusterSpecs.spec.js cypress/integration/
cypress-run --spec "cypress/integration/e2e-createNewCluster.spec.js"
// now get the ISO to the machines and boot them, then run the 2nd 
// script:
cypress-run --spec "cypress/integration/e2e-enterClusterSpecs.spec.js"

### DIFF
--- a/cypress/integration-qe/e2e-createNewCluster.spec.js
+++ b/cypress/integration-qe/e2e-createNewCluster.spec.js
@@ -1,0 +1,17 @@
+import { CLUSTER_NAME, PULL_SECRET, SSH_PUB_KEY, createCluster, generateIso } from './shared';
+
+describe('Flow', () => {
+  it('start from the /clusters page', () => {
+    // Set CYPRESS_BASE_URL environemnt variable
+    // Example: export CYPRESS_BASE_URL=http://localhost:3000
+    cy.visit('');
+  });
+
+  it('create a cluster named ' + CLUSTER_NAME, () => {
+    createCluster(CLUSTER_NAME, PULL_SECRET);
+  });
+
+  it('generate the ISO', () => {
+    generateIso(SSH_PUB_KEY);
+  });
+});
--- a/cypress/integration-qe/e2e-enterClusterSpecs.spec.js
+++ b/cypress/integration-qe/e2e-enterClusterSpecs.spec.js
@@ -1,0 +1,106 @@
+import {
+  CLUSTER_NAME,
+  DNS_DOMAIN_NAME,
+  API_VIP,
+  INGRESS_VIP,
+  NUM_MASTERS,
+  NUM_WORKERS,
+  hostDetailSelector,
+  openCluster,
+  startClusterInstallation,
+  waitForClusterInstallation,
+} from './shared';
+
+import {
+  HOST_REGISTRATION_TIMEOUT,
+  HOST_DISCOVERY_TIMEOUT,
+  VALIDATE_CHANGES_TIMEOUT,
+} from './constants';
+
+describe('Enter cluster details', () => {
+  it('can open the cluster details', () => {
+    openCluster(CLUSTER_NAME);
+  });
+
+  it('wait for the hosts table to be populated', () => {
+    cy.get('table.hosts-table > tbody', { timeout: HOST_REGISTRATION_TIMEOUT }).should(($els) => {
+      expect($els.length).to.be.eq(NUM_MASTERS + NUM_WORKERS);
+    });
+  });
+
+  it('wait for the subnets options to be populated', () => {
+    cy.get('#form-input-hostSubnet-field')
+      .find('option', { timeout: HOST_DISCOVERY_TIMEOUT })
+      .should(($els) => {
+        expect($els.length).to.be.gt(0);
+      })
+      .and(($els) => {
+        expect($els[0]).not.to.have.text('No subnets available');
+      });
+  });
+
+  it('wait for all hosts to reach pending input state', () => {
+    for (let i = 2; i <= NUM_MASTERS + NUM_WORKERS + 1; i++) {
+      cy.contains(hostDetailSelector(i, 'Status'), 'Pending input', {
+        timeout: HOST_DISCOVERY_TIMEOUT,
+      });
+    }
+  });
+
+  it('can set the base domain name', () => {
+    // Cluster configuration - base domain
+    cy.get('#form-input-baseDnsDomain-field').clear();
+    cy.get('#form-input-baseDnsDomain-field').type(DNS_DOMAIN_NAME);
+    cy.get('#form-input-baseDnsDomain-field').should('have.value', DNS_DOMAIN_NAME);
+  });
+
+  it('can select the first subnet CIDR', () => {
+    cy.get('#form-input-hostSubnet-field')
+      .find('option')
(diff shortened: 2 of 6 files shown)


cypress/integration-qe/e2e-createNewCluster.spec.js
```diff
@@ -0,0 +1,57 @@
...
+    // Set CYPRESS_BASE_URL environemnt variable
+    // Example: export CYPRESS_BASE_URL=http://localhost:3000
+    cy.visit('');
+  });
+
+  it('create a cluster named $CLUSTER_NAME', () => {
+    cy.get('#button-create-new-cluster').click();
+    cy.get('#form-input-name-field').should('be.visible');
+    cy.get('#form-input-name-field').clear();
+    cy.get('#form-input-name-field').type(Cypress.env('CLUSTER_NAME'));
```
mareklibra, line 12: I think we should not use `Cypress.env()` directly in the test but copy values to a variable like it is done for `PULL_SECRET` in the `shared.js`.

In addition, we should provide reasonable defaults, if possible (see `cypress.json`).

cypress/integration-qe/e2e-createNewCluster.spec.js
```diff
@@ -0,0 +1,57 @@
+describe('Flow', () => {
+  it('start from the /clusters page', () => {
+    // Set CYPRESS_BASE_URL environemnt variable
+    // Example: export CYPRESS_BASE_URL=http://localhost:3000
+    cy.visit('');
+  });
+
+  it('create a cluster named $CLUSTER_NAME', () => {
```
mareklibra, line 8: Very similar test already exist - `createDummyCluster()` in `shared.js`. Needs to be u[dated for the recently changed UI (pull-secret has been moved). I prefer to reuse&improve such parts as much as possible.

mareklibra:
The flow for Create new cluster has changed in the UI (pull-secret has been moved).

cypress/integration-qe/e2e-createNewCluster.spec.js
ukalifon: The problem with createDummyCluster is that it opens the dialog, and clicks cancel... Then it creates a cluster named "test-infra-cluster" to see that it fails because it expects a cluster with that name to already exist (but it won't fail in our environments)... We need to think together how to share code. I think createDummyCluster should not be in shared.js because it's a very specific test case.

ukalifon: We'll move createDummyCluster out of shared.js and make a more general createCluster function

cypress/integration-qe/e2e-createNewCluster.spec.js
ukalifon: Done

cypress/integration-qe/e2e-enterClusterSpecs.spec.js
```diff
@@ -0,0 +1,98 @@
...
+  waitForClusterInstallation,
+} from './shared';
+
+describe('Enter cluster details', () => {
+  it('can open the cluster details', () => {
+    openCluster(CLUSTER_NAME);
+  });
+
+  it('wait for the hosts table to be populated', () => {
+    cy.get('table.hosts-table > tbody', { timeout: 10 * 60 * 1000 }).should(($els) => {
```
mareklibra, line 20: nit: Can we put the timeout value to a constant and co-locate them? Please consider even introducing a new `constants.js` file for such things.

cypress/integration/shared.js
```diff
@@ -99,3 +168,18 @@ export const checkValidationMessage = (cy, expectedMsg) => {
...
+    expect($elem).to.be.enabled;
+  });
+  cy.get('button[name="install"]').click();
+  // wait for the progress description to say "Installing"
+  cy.contains('div.pf-c-progress__description', 'Installing', { timeout: 60 * 1000 });
+};
+
+export const waitForClusterInstallation = () => {
+  // wait up to 1 hour for the progress description to say "Installed"
+  cy.contains('div.pf-c-progress__description', 'Installed', { timeout: 60 * 60 * 1000 });
```
mareklibra, line 184: Let's co-locate all explicitly set timeouts.

cypress/integration/shared.js
```diff
@@ -17,8 +17,18 @@ export const clusterNameLinkSelector = '[data-label="Name"] > a'; // on '/cluste
...
+  // NOTE: The first row is number 2! Shift your indexes...
+  `table > tbody:nth-child(${row}) > tr:nth-child(1) > [data-label="${label}"]`;
 
 export const PULL_SECRET = Cypress.env('PULL_SECRET');
+export const SSH_PUB_KEY = Cypress.env('SSH_PUB_KEY');
+export const CLUSTER_NAME = Cypress.env('CLUSTER_NAME');
+export const DNS_DOMAIN_NAME = Cypress.env('DNS_DOMAIN_NAME');
+export const API_VIP = Cypress.env('API_VIP');
+export const INGRESS_VIP = Cypress.env('INGRESS_VIP');
+export const NUM_MASTERS = parseInt(Cypress.env('NUM_MASTERS'));
```
mareklibra, line 30: Please ad reasonable defaults to `/cypress.json`, otherwise this is `NaN`

cypress/integration-qe/e2e-createNewCluster.spec.js
ukalifon: For now we'll leave createDummyCluster in shared.js.

cypress/integration-qe/e2e-enterClusterSpecs.spec.js
ukalifon: Added constants.js and put the timeouts there

cypress/integration/shared.js
mareklibra: Related: https://github.com/openshift-metal3/facet/pull/269

mareklibra:
Please rebase on top of https://github.com/openshift-metal3/facet/pull/269 .
Once the numeric timeout constants are co-located, this PR lgtm. Thank you!

cypress/integration/shared.js
ukalifon: Done

cypress/integration/shared.js
ukalifon: Done

PR #260, opened 2020-07-26
Raz/MGMT-1231

RazRegev:
If user pass namespace argument update the default inventory-url to use bm-inventory under this namespace.

### DIFF
--- a/deploy/deploy_config.sh
+++ b/deploy/deploy_config.sh
@@ -7,7 +7,7 @@ usage() {
   echo "Usage: $0 [ -t deploy-confing-template ] [ -u BM-INVENTORY-URL ] [ -i UI_IMAGE_NAME ] [ -n NAMESPACE ]" 1>&2
 }
 
-BM_INVENTORY_URL="http://bm-inventory.assisted-installer.svc.cluster.local:8090"
+BM_INVENTORY_URL="http://bm-inventory.__NAMESPACE__.svc.cluster.local:8090"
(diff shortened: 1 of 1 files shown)


mareklibra:
Thank you for your patch. Honestly, I like approach from #258 most.

Next time you can try `git push --force` to your branch when a PR is open here. Just prefix the PR title with `WIP` to avoid any confusions while you keep working on it.

RazRegev:
> Thank you for your patch. Honestly, I like approach from #258 most.
> 
> Next time you can try `git push --force` to your branch when a PR is open here. Just prefix the PR title with `WIP` to avoid any confusions while you keep working on it.

Right, will do next time. 
Why would you like #258 more then this one?
I think that it is more organized like this but I can also reopen #258

mareklibra:
> Why would you like #258 more then this one?

Code-style reasons only. Two people, two opinions. In that case the function is never called twice and there is expectation that this won't change. The logic fits one line only. It is called directly from already "more" complex command decreasing readability. If you use the simple setting of env variable, the code readability is increased - a future maintainer will not need to "jump" but just read the script from top to bottom.

RazRegev:
> > Why would you like #258 more then this one?
> 
> Code-style reasons only. Two people, two opinions. In that case the function is never called twice and there is expectation that this won't change. The logic fits one line only. It is called directly from already "more" complex command decreasing readability. If you use the simple setting of env variable, the code readability is increased - a future maintainer will not need to "jump" but just read the script from top to bottom.

Legit, fixed that.

mareklibra:
Thanks!

PR #265, opened 2020-07-30
Sync clusterDetail tests

mareklibra:


cypress/integration/clusterDetail.spec.js
```diff
@@ -306,8 +308,9 @@ describe('Cluster Detail', () => {
     cy.get('.pf-c-modal-box__title').contains('Cluster Events');
     cy.get('.pf-c-table').find('tr').should('have.length.greaterThan', 0);
     // this fails... I tried with should('have.text', 'Registered cluster "test-infra-cluster"') but it also fails
```
ukalifon, line 310: You should remove my comment "this fails..."

PR #285, opened 2020-08-21
HLD Automation

ukalifon:
First batch of HLD test plan test cases


jkilzi:
@ukalifon I think this PR is too old and might be already implemented downstream, is it still necessary?

PR #293, opened 2020-08-30
E2E flow is broken due to UI changes. 

lalon4:
Fixed createCluster & generateIso methods, upon UI changes

### DIFF
--- a/cypress/integration/shared.js
+++ b/cypress/integration/shared.js
@@ -66,7 +66,7 @@ export const createCluster = (clusterName, pullSecret) => {
   // feed in the pull secret
   cy.get('#form-input-pullSecret-field').clear();
   pasteText(cy, '#form-input-pullSecret-field', pullSecret);
-  cy.get('form').submit();
+  cy.get('button[name="save"]').click();
   cy.get('#button-download-discovery-iso').should('be.visible');
   cy.get('#form-input-name-field').should('have.value', clusterName);
 };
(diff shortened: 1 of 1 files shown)


cypress/integration/shared.js
```diff
@@ -114,17 +114,18 @@ export const generateIso = (sshPubKey) => {
   // click to download the discovery iso
   cy.get('#button-download-discovery-iso').click();
   // see that the modal popped up
-  cy.get('h1#pf-modal-part-8').should('be.visible');
+  cy.get('#pf-modal-part-6').should('be.visible');
   // feed in the public ssh key
-  pasteText(cy, '#form-input-sshPublicKey-discovery-field', sshPubKey);
```
nshidlin, line 119: Why do you use type() instead of the pasteText method? Does this work reliably?

cypress/integration/shared.js
```diff
@@ -114,17 +114,18 @@ export const generateIso = (sshPubKey) => {
...
+  cy.get('#form-input-sshPublicKey-discovery-field').type(sshPubKey);
   let aborted = false;
   cy.server({
     onAnyAbort: (...args) => {
       aborted = true;
       console.log('-- onAnyAbort: ', ...args);
     },
   });
-  cy.get('.pf-c-modal-box__footer > .pf-m-primary').contains('Get Discovery ISO');
+  cy.get('.pf-c-modal-box__footer > .pf-m-primary').should('be.visible');
```
nshidlin, line 127: having both should('be.visible') and contains('Generate Discovery ISO') seems redundant; wdyt?

cypress/integration/shared.js
lalon4: From some reason, after using this pasteText method (only here), outputs an empty field, and cause a failure. 
Started to look into it and it has issue with the {backspace}, although in all other areas when using this method it is working fine. Didnt want to waist more time to investigate as i assume the UI will change, again. WDYT?

cypress/integration/shared.js
lalon4: be.visible means that the button is clickable where the contains only check its text, correct me if im wrong

cypress/integration/shared.js
nshidlin: I see what you mean. Might due to the new ability to drag a file in.

cypress/integration/shared.js
nshidlin: If the element is not visible the contains('Generate Discovery ISO') will fail. But failing on should('be.visible') gives a better indication of what failed.

ukalifon:
/lgtm
Please merge it and check that you can deploy with the job using the web-ui flow

nshidlin (approved):

PR #295, opened 2020-09-02
Fix ISO download with Cypress

cdvultur:
Fix for the Cypress ISO generation and Download

### DIFF
--- a/cypress/integration/shared.js
+++ b/cypress/integration/shared.js
@@ -135,7 +135,7 @@ export const generateIso = (sshPubKey) => {
     if (aborted) {
       cy.log('Long-running XHR was aborted');
       cy.get('.pf-c-alert').contains('Failed to download');
-      cy.get('.pf-c-modal-box__footer > .pf-m-primary').contains('Get Discovery ISO', {
+      cy.get('.pf-c-modal-box__footer > .pf-m-primary').contains('Generate Discovery ISO', {
         timeout: 5 * 60 * 1000,
       });
       cy.get('.pf-c-modal-box__footer > .pf-m-primary').click();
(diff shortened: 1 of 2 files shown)


cypress/integration/shared.js
```diff
@@ -158,6 +158,9 @@ export const downloadFileWithChrome = (downloadButton, resultantFilename) => {
   cy.exec(`[ -f ${resultantFilename} ] && rm -rf ${resultantFilename}`, {
     failOnNonZeroExit: false,
   });
+  cy.wait(2 * 60 * 1000);
+  cy.get('.pf-c-modal-box__footer > .pf-m-primary').click();
+  cy.get(downloadButton, { timeout: 2 * 60 * 1000 }).contains('Download Discovery ISO');
```
mareklibra, line 163: I think this is too specific for this generic function. From the name, the function can be used to handle other file downloads too.

cypress/integration/shared.js
```diff
@@ -158,6 +158,9 @@ export const downloadFileWithChrome = (downloadButton, resultantFilename) => {
   cy.exec(`[ -f ${resultantFilename} ] && rm -rf ${resultantFilename}`, {
     failOnNonZeroExit: false,
   });
+  cy.wait(2 * 60 * 1000);
```
mareklibra, line 161: Can we add a condition instead of the delay? If not, please write add a comment describing what do we wait for and preferably verify after this `wait()`.

cypress/integration/shared.js
cdvultur: In the end I've removed the lines from there and added sleep to the parent caller file. 
There is some bug in there where while being called via Cypress, 2 images are created at the same time.

cypress/integration/shared.js
cdvultur: Currently the wait is a ruff estimate of what is happening in the backend. Nothing in UI to base the wait on. Still this is moved to the caller parent file.

mareklibra (approved):

PR #296, opened 2020-09-03
Downgrade from Yarn to NPM

mareklibra:


jkilzi:
@mareklibra  after @rawagner new PR I think this is no longer necessary, right?

PR #309, opened 2020-09-23
Webui on stage

cdvultur:
This is a follow up of the https://github.com/openshift-metal3/facet/pull/284 , which was further adapted for both Hosted and local deployments .
Tested against both Hosted on  v1.0.43  and PSI on 1.0.37.

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -10,9 +10,12 @@
     "INGRESS_VIP": "192.168.126.101",
     "NUM_MASTERS": 3,
     "NUM_WORKERS": 0,
-    "API_BASE_URL": "",
+    "API_BASE_URL": "http://api.example.com",
     "INTEGRATION_ENV": "",
     "OCM_USER": "",
-    "OCM_USER_PASSWORD": ""
+    "OCM_USER_PASSWORD": "",
+    "OCM_COOKIE_NAME": "cs_jwt",
+    "OCM_TOKEN_DEST": "~/Downloads/token",
+    "ISO_PATTERN": "~/Downloads/discovery-image-*.iso"
   }
 }
--- a/cypress.json
+++ b/cypress.json
@@ -9,6 +9,10 @@
     "API_VIP": "192.168.126.100",
     "INGRESS_VIP": "192.168.126.101",
     "NUM_MASTERS": 3,
-    "NUM_WORKERS": 0
+    "NUM_WORKERS": 0,
+    "API_BASE_URL": "",
+    "INTEGRATION_ENV": "",
+    "OCM_USER": "",
+    "OCM_USER_PASSWORD": ""
   }
 }
--- a/cypress/integration-qe/basicUserAuthentication.spec.js
+++ b/cypress/integration-qe/basicUserAuthentication.spec.js
@@ -1,0 +1,18 @@
+import { verifyPullSecretHosted } from './ocmShared';
+
+describe('test basic user authentication', () => {
+  it('navigate from openshift portal to assisted installer', () => {
+    cy.visit('');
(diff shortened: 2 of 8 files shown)


cypress/integration-qe/basicUserAuthentication.spec.js
```diff
@@ -0,0 +1,18 @@
+import { verifyPullSecretHosted } from './ocmShared';
+
+describe('test basic user authentication', () => {
+  it('navigate from openshift portal to assisted installer', () => {
+    cy.visit('');
+    cy.get('[aria-current="page"]').contains('Clusters').click();
+    cy.get('button').contains('Create cluster').click();
```
mareklibra, line 7: nit: to make it robust, let's do not assume that there is just single `<button>` on the page

cypress.json
```diff
@@ -9,6 +9,18 @@
...
     "NUM_MASTERS": 3,
-    "NUM_WORKERS": 0
+    "NUM_WORKERS": 0,
+    "API_BASE_URL": "",
+    "INTEGRATION_ENV": "",
+    "OCM_USER": "",
+    "OCM_USER_PASSWORD": "",
+    "OCM_COOKIE_NAME": "",
+    "OCM_TOKEN_DEST": "",
+    "ISO_PATTERN": ""
```
mareklibra, line 19: What about adding a dummy default value here?

cypress/integration-qe/e2e-enterClusterSpecs.spec.js
```diff
@@ -58,7 +68,7 @@ describe('Enter cluster details', () => {
   it('can select the first subnet CIDR', () => {
     cy.get('#form-input-hostSubnet-field')
       .find('option')
-      .then(($els) => $els.get(0).setAttribute('selected', 'selected'))
+      .then(($els) => $els.get(1).setAttribute('selected', 'selected'))
```
mareklibra, line 71: Such approach works but it is low-level with assumptions about implementation.

It is not a must, but with cypress we can mimic user's clicking on elements.
I guess it will be result in more code but closer to reality.

cypress/integration-qe/e2e-enterClusterSpecs.spec.js
```diff
@@ -115,3 +125,13 @@ describe('Donwload kubeconfig', () => {
     cy.exec('mv -f ~/Downloads/kubeconfig ~');
   });
 });
+
+describe('Refresh cookie on disk', () => {
+  it('refresh cookie', () => {
```
mareklibra, line 130: I can see similar case in `e2e-createNewCluster.spec.js`. I guess, motivation behind both of them is the same. Can we unify?

And is this really a test case/scenario? If not, what about using hooks like `beforeEach()` or similar?

cypress/integration/ocmShared.js
```diff
@@ -0,0 +1,77 @@
...
+
+  cy.get('#button-create-new-cluster', {
+    timeout: DEFAULT_API_REQUEST_TIMEOUT,
+  });
+};
+
+export const writeCookieToDisk = async (cookiename, cookiefile) => {
+  cy.getCookie(cookiename).then((cookie) => {
+    // dumping cookie to disk
+    cy.exec(`rm -f ${cookiefile}; echo ${cookie.value} >> ${cookiefile}`);
```
mareklibra, line 33: nit: with replacing `>>` by single `>` you can skip the `rm`

mareklibra: The `cookie.value` can theoretically contain. Wrapping in single quotes might help.

ukalifon (approved):
/lgtm

jtomasek (approved):

PR #312, opened 2020-09-29
Deployment types

ukalifon:
Adding support for DHCP-VIP deployments, as well as advanced networking
deployments. Moving the install cluster steps to a separate script so
that it will be possible to run the additional scripts for the above
deployment types before launching the installation.


cdvultur (approved):
/lgtm

jtomasek (approved):

PR #314, opened 2020-10-02
Fail faster

ukalifon:
This change will fail if the cluster installation didn't start, and
won't wait for it to finish even if it didn't start. It also waits
less for files to download, and only waits long for the ISO file
download.


yobshans (approved):

yobshans:
looks like ok

PR #315, opened 2020-10-05
Install preparation timeout

ukalifon:
Increasing the timeout for the cluster state to switch from "preparing
to install" to "installing". It seems that 1 minute was not enough.


cdvultur (approved):
/lgtm

PR #316, opened 2020-10-07
Timeout for start installation

ukalifon:
Increasing the timeout for the "Install Cluster" button to become
enabled. When DHCP is enabled is seems to take more time for the
button to be clickable, so increasing the timeout to 1 minute.


cdvultur (approved):
/lgtm

mareklibra:
IIRC, backend has internally 2 minutes timeout. I would be even more generous on frontend.

PR #319, opened 2020-10-12
Adjust to the new search input id

jtomasek:
Depends on https://github.com/mareklibra/facet-lib/pull/255


jkilzi:
@jtomasek do we still need this?

jkilzi:
Closing since Cypress tests were moved to downstream

PR #321, opened 2020-10-12
Updates to the shared infrastructure

ukalifon:
Updates to test-infra

Fixing the tests for test-infra, and unifying the code used by test-
infra and the QEs.

Changes to createCluster and createDummyCluster:
1) createDummyCluster will feed in the cluster name and pull secret,
   and will leave you in the create cluster screen.
2) createDummyCluster will not finish creating the cluster, and the
   user will have the option to call cancelDummyCluster or complete
   the creation process.
3) createCluster uses createDummyCluster in order to prevent code
   duplication. Most scenarios will call createCluster.

### DIFF
--- a/cypress/integration/shared.js
+++ b/cypress/integration/shared.js
@@ -9,11 +9,11 @@ import {
   START_INSTALLATION_TIMEOUT,
 } from './constants';
 
-export const testInfraClusterName = 'test-infra-cluster';
+export const testInfraClusterName = 'test-infra-cluster-assisted-installer';
 export const testInfraClusterHostnames = [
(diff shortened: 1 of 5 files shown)


cdvultur (approved):
/lgtm

cypress/integration/shared.js
```diff
@@ -351,6 +336,28 @@ export const getDhcpVipState = (cy) => {
   });
 };
 
+export const getClusterState = (cy) => {
```
mareklibra, line 339: good idea

mareklibra (approved):

PR #324, opened 2020-10-15
Increate cluster creation timeout

ukalifon:
Due to changes in the installer - it now waits for the cluster
version to be up, so installation takes much longer. This patch
increases the timeout.


cdvultur (approved):
/lgtm

PR #325, opened 2020-10-19
adding test for dynamic DNS service; also separated the input of DNS …

cdvultur:
Adding support for dynamic dns service. 
Also adapted the rest of the test to make fit for the call of this test

### DIFF
--- a/cypress/integration/e2e-setClusterDnsDomain.spec.js
+++ b/cypress/integration/e2e-setClusterDnsDomain.spec.js
@@ -1,0 +1,21 @@
+import {
+  CLUSTER_NAME,
+  openCluster,
+  saveClusterDetails,
+  setClusterDnsDomain,
+  DNS_DOMAIN_NAME,
+} from './shared';
+
+describe('Set DNS Domain -> ', () => {
+  it('Open the cluster details', () => {
+    openCluster(CLUSTER_NAME);
+  });
+
+  it('Set DNS Domain', () => {
+    setClusterDnsDomain(DNS_DOMAIN_NAME);
+  });
+
+  it('Save changes', () => {
+    saveClusterDetails(cy);
+  });
+});
--- a/cypress/integration/shared.js
+++ b/cypress/integration/shared.js
@@ -8,6 +8,7 @@ import {
   FILE_DOWNLOAD_TIMEOUT,
   START_INSTALLATION_TIMEOUT,
 } from './constants';
+import { resolvePlugin } from '@babel/core';
 
 export const testInfraClusterName = 'test-infra-cluster-assisted-installer';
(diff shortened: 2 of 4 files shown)


cypress/integration/shared.js
```diff
@@ -451,6 +451,12 @@ export const saveClusterDetails = (cy) => {
   // click the 'save' button in order to save changes in the cluster info
   cy.get('button[name="save"]', { timeout: VALIDATE_CHANGES_TIMEOUT }).should('be.enabled');
   cy.get('button[name="save"]').click();
+  cy.wait(2 * 1000);
```
mareklibra, line 454: It would be better to block on an UI change instead of dummy waiting. But feel free to ignore this comment in case it would be too complicated in this case.

Just please be careful in case of potential subsequent checks, i.e. whether there are/arn't errors rendered - 2 seconds might not be enough and the test will miss errors ... Such thinking would be easier if you block processing on an UI change ^^^.

cypress/integration/e2e-setClusterDnsDomain.spec.js
```diff
@@ -0,0 +1,21 @@
...
+  it('Open the cluster details', () => {
+    openCluster(CLUSTER_NAME);
+  });
+
+  it('Set DNS Domain', () => {
+    setClusterDnsDomain(DNS_DOMAIN_NAME);
+  });
+
+  it('Save changes', () => {
+    saveClusterDetails(cy);
```
mareklibra, line 19: Might be good to check list of errors (either expected ones or none displayed once it is clear that response from API is received).

cypress/integration/shared.js
```diff
@@ -451,6 +451,12 @@ export const saveClusterDetails = (cy) => {
...
+};
+
+export const enableRoute53 = (cy) => {
+  cy.get('#form-input-useRedHatDnsService-field').click();
+  cy.get('#form-input-baseDnsDomain-field').contains('route53');
```
nshidlin, line 459: Maybe rather than checking for route53 hard coded you can make a GET request to /domains and compare to the response in case route53 is changed in the future

cdvultur: Refactored the check

cypress/integration/e2e-setClusterDnsDomain.spec.js
cdvultur: As discussed the list is there only on error and not on happy flow

cypress/integration/shared.js
cdvultur: As discusses only on error we have the error list, and there are no IDs to match on successful save.

mareklibra (approved):

PR #326, opened 2020-10-19
paramatizing generateIso to allow passing proxy params

nshidlin:
This PR parameterizes generateIso to allow passing proxy params
Filling of proxy params is done by  setProxyValues in shared.js


cdvultur (approved):
/lgtm

PR #328, opened 2020-10-22
increating ISO download timeout

cdvultur:


cypress/integration/constants.js
```diff
@@ -15,6 +15,6 @@ export const GENERATE_ISO_TIMEOUT = 2 * 60 * 1000;
 // timeout for downloading files
 export const FILE_DOWNLOAD_TIMEOUT = 60 * 1000;
 // timeout for downloading the ISO image
-export const ISO_DOWNLOAD_TIMEOUT = 30 * 60 * 1000;
+export const ISO_DOWNLOAD_TIMEOUT = 60 * 60 * 1000;
```
mareklibra, line 18: Such long timeout might become bottleneck, especially when waiting in an error-state and not just for slow download. Let's see how it behaves.

mareklibra (approved):

PR #331, opened 2020-10-28
Adjusting the index in the list.

achuzhoy:
There was a change in UI and now the cluster
name appears as item #3 instead of #2. Adjusting
the automation accordingly.

Signed-off-by: Alexander Chuzhoy <achuzhoy@redhat.com>


achuzhoy:
Passed in test against PSI.

PR #342, opened 2020-11-17
Adjusting automation with UI changes.

achuzhoy:
Seems like the ID has changed for the ISO
download widget. Adjusting.

Signed-off-by: Alexander Chuzhoy <achuzhoy@redhat.com>


achuzhoy:
Not needed - was testing a wrong portal

PR #374, opened 2020-12-21
 Increated timeouts

cdvultur:
Increasing timeout


achuzhoy:
@cdvultur since DEFAULT_API_REQUEST_TIMEOUT is used in several places - could it be a real bug when the timeout takes too long to answer.

cdvultur:
@achuzhoy Added the DEFAULT_CREATE_CLUSTER_BUTTON_SHOW_TIMEOUT to be more specific and not collide with the API ones.

nshidlin (approved):

PR #381, opened 2021-01-11
Doing cleanup

cdvultur:
Doing cleanup

PR #382, opened 2021-01-11
Updated upstream tests

mareklibra:
Scope of this PR has changed to
- refactor `shared.js` code into `shared` subfolder
- use Typescript in tests
- update upstream tests
- update documentation
- update Cypress to latest
- replace deprecated `cy.router` by recent interceptors

The initial state for the upstream tests is the result of https://github.com/openshift/assisted-test-infra.

Related `yarn` scripts:
-    "cypress-open": to test against real backend with handy UI (most common while developing)
-    "cypress-run": real backend but headless

Downstream tests need to be updated to:
- `import` from the new `shared` folder instead of former `shared.js`. For limitted time, there is `shared-deprecated.js` file with former content.
- rename source files from `*.js` to `*.ts` 
- add/fix Typescript types 
- replace `cy.router` by new interceptor API (if any - upstream is updated by this PR)
---

At the beginning of this effort, we hoped in having automatically updated mock environment to execute tests by recording/replaying network communication. Unfortunately, such environment showed to be flaky, unless timing of the requests gets aligned between record- and replay-time.

There are still more ideas to explore towards having such mocked environment but it was decided to split the work and deal with it in a follow-up. One of the promising ideas is to decorate `Cypress.cy.get()` to record time-like information which the replay can sync on later.

---

TODO:
- [x] rebase on top of "recently pruned" tests - #381
- [x] use Typescript in tests
- [x] test with POST, PATCH
- [x] cover all existing upstream test suites
- [x] update existing upstream code to the new Cypress (i.e. use interceptors in hacks, including downstream)
- [x]  write scenario-based documentation
- [x] build container images at every PR (the only tag is git sha, i.e. the `latest` or `stable` are not touched)
- [x] https://github.com/openshift-assisted/assisted-ui-lib/pull/411


package.json
```diff
@@ -78,10 +80,10 @@
     "@babel/node": "^7.2.2",
     "@babel/preset-env": "^7.3.4",
     "@testing-library/dom": "^7.11.0",
-    "cypress": "^4.5.0",
+    "cypress": "^6.2.1",
```
cdvultur, line 83: This will break some of the things we run in QE, since we share the same package json. Still this will just impact the phase1 job

mareklibra: It's on the TODO "update existing code to the new Cypress". I believe, we will need to especially move from the use of `route()` to interceptors there.

cdvultur:
/lgtm

Thanks for the effort of getting this done

mareklibra:
Rebased.
Verified that all tests are still passing.

ukalifon (approved):
/lgtm

jtomasek (approved):

PR #388, opened 2021-01-28
Converging QE and DEV code

ukalifon:
First batch of changes needed to run the QE's tests with the
upstream infrastructure.

### DIFF
--- a/cypress/integration/shared/clusterConfiguration.ts
+++ b/cypress/integration/shared/clusterConfiguration.ts
@@ -2,7 +2,12 @@ import {
   DEFAULT_SAVE_BUTTON_TIMEOUT,
   HOST_REGISTRATION_TIMEOUT,
   VALIDATE_CHANGES_TIMEOUT,
+  HOSTS_DISCOVERY_TIMEOUT,
+  HOST_DISCOVERY_TIMEOUT,
 } from './constants';
+
+import { makeApiCall, clusterIdFromUrl } from './common';
+
 import { DNS_DOMAIN_NAME, NUM_MASTERS, NUM_WORKERS } from './variables';
 
 export const hostDetailSelector = (shiftedRowIndex: number, label: string) =>
@@ -78,6 +83,40 @@ export const waitForHostTablePopulation = (
   });
 };
 
+export const waitForHostsSubnet = (cy: Cypress.cy) => {
+  // wait until hosts subnet populated in the cluster details
+  cy.get('#form-input-hostSubnet-field')
+    .find('option', { timeout: HOST_DISCOVERY_TIMEOUT })
+    .should(($els) => {
+      expect($els.length).to.be.gt(0);
+    })
+    .and(($els) => {
+      expect($els[0]).not.to.have.text('No subnets available');
+    });
+};
+
+export const waitForHostsToBeReady = (
+  cy: Cypress.cy,
+  numMasters = NUM_MASTERS,
+  numWorkers = NUM_WORKERS,
+) => {
+  // wait until hosts are getting to pending input state
+  for (let i = 2; i <= numMasters + numWorkers + 1; i++) {
+    cy.contains(hostDetailSelector(i, 'Status'), 'Ready', {
+      timeout: HOSTS_DISCOVERY_TIMEOUT,
+    });
+  }
+};
+
+export const setClusterSubnetCidr = (cy: Cypress.cy) => {
+  // select the first subnet from list
+  cy.get('#form-input-hostSubnet-field')
+    .find('option')
+    .then(($els) => $els.get(1).setAttribute('selected', 'selected'))
+    .parent()
+    .trigger('change');
+};
+
 export const setHostsRole = (
   cy: Cypress.cy,
   masterHostnamePrefix: string,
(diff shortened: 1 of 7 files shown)


cypress/integration/shared/clusterConfiguration.ts
```diff
@@ -78,6 +83,41 @@ export const waitForHostTablePopulation = (
   });
 };
 
+export const waitForHostsSubnet = (cy) => {
```
mareklibra, line 86: Please note we moved from plain JavaScript to TypeScript.
At the build type, static type-analysis is done. To make it possible, we need to specify types whenever they can not be deduced from context.

```suggestion
export const waitForHostsSubnet = (cy: Cypress.cy) => {
```

cypress/integration/shared/clusterConfiguration.ts
```diff
@@ -78,6 +83,41 @@ export const waitForHostTablePopulation = (
...
+    .find('option', { timeout: HOST_DISCOVERY_TIMEOUT })
+    .should(($els) => {
+      expect($els.length).to.be.gt(0);
+    })
+    .and(($els) => {
+      expect($els[0]).not.to.have.text('No subnets available');
+    });
+};
+
+export const waitForHostsToBeReady = (cy, numMasters = NUM_MASTERS, numWorkers = NUM_WORKERS) => {
```
mareklibra, line 98: `cy` is missing type
`numMasters` type can be automatically deduced from the initial value (means `number` in this case)

cypress/integration/shared/common.ts
```diff
@@ -78,3 +79,103 @@ export const waitForClusterState = (cy: Cypress.cy, desiredState: string, retrie
     }
   });
 };
+
+export const setProxyValues = (httpProxy = null, httpsProxy = null, noProxy = null) => {
```
mareklibra, line 83: Since this is related to a specific page (Cluster configuration in this case), then it should not be in the `common.ts` file

cypress/integration/shared/variables.ts
```diff
@@ -1,6 +1,7 @@
 export const PULL_SECRET = Cypress.env('PULL_SECRET');
 export const SSH_PUB_KEY = Cypress.env('SSH_PUB_KEY');
 export const CLUSTER_NAME = Cypress.env('CLUSTER_NAME');
+export const CLUSTER_BASE_DOMAIN = Cypress.env('CLUSTER_BASE_DOMAIN');
```
mareklibra, line 4: Dupplicate of `DNS_DOMAIN_NAME`. I think we should unify all code to either of those two.

mareklibra (changes requested):

hacks/cypress_env_local.sh
```diff
@@ -14,10 +14,12 @@ export CYPRESS_BASE_URL=http://localhost:3000
 export CYPRESS_SSH_PUB_KEY=$(cat ~/.ssh/id_rsa.pub)
 
 export CYPRESS_CLUSTER_NAME=test-infra-cluster-assisted-installer
+export CYPRESS_CLUSTER_BASE_DOMAIN='redhat.com'
 export CYPRESS_CLUSTER_HOSTNAME_MASTER_PREFIX=test-infra-cluster-assisted-installer-master
 export CYPRESS_CLUSTER_HOSTNAME_WORKER_PREFIX=test-infra-cluster-assisted-installer-worker
 export CYPRESS_PULL_SECRET=$(cat ~/pull-secret)
 export CYPRESS_ISO_PATTERN="~/Downloads/cluster-*-discovery.iso"
+export CYPRESS_OPENSHIFT_VERSION='OpenShift 4.6.8'
```
cdvultur, line 22: You could use just '4.7', so that we pass the value not the display value. Also that would spare us on updating on each new release.

cypress/integration/shared/common.ts
ukalifon: setProxyValues is required by generateIso, and they are both in common.ts. It would not be proper to put it in clusterConfiguration (almost everything is on that page).

cypress/integration/shared/clusterConfiguration.ts
ukalifon: Done.

cypress/integration/shared/clusterConfiguration.ts
ukalifon: Done

cypress/integration/shared/variables.ts
ukalifon: Good catch. I'll change to DNS_DOMAIN_NAME.

hacks/cypress_env_local.sh
ukalifon: Done. Aligned with your code from PR 96 downstream.

cypress/integration/shared/clusterConfiguration.ts
```diff
@@ -100,3 +139,117 @@ export const setHostsRole = (
...
+      const readDhcpAllocation = (response) => {
+        resolve(response.body.vip_dhcp_allocation);
+      };
+
+      makeApiCall(`/api/assisted-install/v1/clusters/${id}`, 'GET', readDhcpAllocation);
+    });
+  });
+};
+
+export const disableDhcpVip = (cy: Cypress.cy, apiVip = null, ingressVip = null) => {
```
mareklibra, line 155: nitpick:
```
apiVip?: string = null, 
```
or better (considering context):
```
apiVip = '',
```

mareklibra (approved):

PR #399, opened 2021-02-09
add cypress-wait-until package

hardengl:


hardengl:
@mareklibra @cdvultur I see many uses for this package in our current automation, let me know what y'all think.

mareklibra:
Thanks for the contribution. I am ok with it as long as we will not misuse it. I would like to avoid heavy use of it on places where regular Cypress mechanisms can be used. Cases like chained waiting for multiple conditions make totally sense to me.

Can you please add typescript configuration? I saw that extra steps are needed.

hardengl:
I will update the ts config. Yes, I agree to only use where needed.

mareklibra (approved):

PR #402, opened 2021-02-11
add selector: byDataTestId

hardengl:
adding selectors.ts file to hold cypress command selectors. @mareklibra please let me know your thoughts.


cypress/support/selectors.ts
```diff
@@ -0,0 +1,10 @@
+// eslint-disable-next-line @typescript-eslint/no-namespace
+declare namespace Cypress {
+  interface Chainable<Subject> {
+    byDataTestID(selector: string): Chainable<Element>;
+  }
+}
+
+Cypress.Commands.add('byDataTestID', (selector: string) => {
```
mareklibra, line 8: nit: 
```suggestion
Cypress.Commands.add('byDataTestID', (dataTestId: string) => {
```

mareklibra:
> please let me know your thoughts

Considering this particular case, such approach looks overwhelming to me, I do not see much benefits of it. Looks to me like just indirection which reduces maintainability by new team members. I would have different opinion in case of more complex logic inside the command. Please correct me.

hardengl:
https://docs.cypress.io/guides/references/best-practices.html#Real-World-Example

mareklibra:
Thanks for the link. I miss a list of benefits for this simple case which we have in this PR. Can you please elaborate?

hardengl:
Main reason is to write our code base with respect to cypress best practices. This is just a taste of what should be, in practice all of our selectors should be in support/selectors.ts file. Then most of the other functions that are currently in shared/*.ts should either be moved to their respective file under support folder or to a views folder that can hold functions for the different views. IMO this makes it much easier for development. In addition, it will make our cypress test code much easier to follow.

mareklibra:
If we move all selectors to one pile. they will become very hard to maintain. We will need to separate them anyway. My preference is in doing that by page or similar logical UI group. This remark is valid for both cases - either `shared/*` or `selectors`.

It's good to understand that recent proposal is just "a taste". IMO, this approach makes sense only if it is wrapping a general-purpose/abstracted pattern. Simple one-liners without additional logic are not that case. However, by having more of those useful ones, I agree it makes sense to start using them even for that simple byTestDataId case, just because of consistency.

Generally speaking, every abstraction or indirection is source of complications for new-comers to the project. It is good to use it just if needed, but not blindly. In that particular case, it is even both hard to use git grep and the IDEs have hard time to navigate the developer to the "command" implementation.

So my proposal is to flip the effort - to start looking for blocks of code where such refactoring will bring clear benefit and evolve from there.

hardengl:
In theory, we shouldn't have many selectors if our html elements are labeled as they should be for testing purposes. By default Cypress will automatically prefer elements with attributes in this order:

data-cy
data-test
data-testid
id
class
tag
attributes
nth-child

For ease of testing, we could start to label all html elements of concern with the 'data-cy' or 'data-test' attribute. This will make the code base easier to read and follow and will also optimize our implementation to fit what the Cypress framework expects and the documentation recommends.

hardengl:
This is the structure I propose, and the structure that follows Cypress recommended best practice. If there is a good reason to not follow best practice for this use case please let me know. It really isn't that far off from what you currently have. 

└──cypress
    ├── support    <--- add commands to Cypress 'cy.' global, other support configurations
    │   ├── commands.js
    │   ├── index.js
    │   ├── login.js
    │   └── selectors.js
    ├── fixtures                <--- mock data
    │   └── example.json
    ├── plugins
    │   └── index.js            <--- webpack-preprocessor, environment variables, baseUrl, custom tasks
    ├── integration             <--- test suites
    │   ├── tests.spec.js
    │   
    └── views                   <--- helper objects containing assertions and commands
        ├── ClusterListPage.ts

jkilzi:
@mareklibra @hardengl do we still want/need this PR?

hardengl:
closed

PR #416, opened 2021-03-04
Updating column name in Cypress shared code

ukalifon:
The hosts table column is now called 'Hardware Status' instead of
'Status'.
NOTE: this change should only go in with the wizard branch.


mareklibra:
@ukalifon , is this still needed? https://github.com/openshift-assisted/assisted-ui-lib/pull/506 has been merged.

ukalifon:
This is not needed, the column name didn't change.

PR #443, opened 2021-06-04
Update deps and config to transpile non-bundled ui lib

rawagner:


package.json
```diff
@@ -8,43 +8,40 @@
   },
```
rawagner, line 8: changes here are from updated `react-scripts`.
I've also moved some dependencies to `devDependencies`

src/styles/index.css
```diff
@@ -6,4 +6,3 @@
 */
 @import '~@patternfly/patternfly/patternfly.css';
 @import '~@patternfly/patternfly/patternfly-addons.css';
-@import '~openshift-assisted-ui-lib/dist/index.css';
```
rawagner, line 9: since lib is not bundled anymore, there is no 'main' css

jtomasek (approved):

jkilzi (approved):

PR #456, opened 2021-07-13
Adapt to new assisted-ui-lib structure

mareklibra:
Depends on:
- [x] https://github.com/openshift-assisted/assisted-ui-lib/pull/681

### DIFF
--- a/src/styles/index.css
+++ b/src/styles/index.css
@@ -6,4 +6,3 @@
 */
 @import '~@patternfly/patternfly/patternfly.css';
 @import '~@patternfly/patternfly/patternfly-addons.css';
-@import '~openshift-assisted-ui-lib/dist/index-ocm.css';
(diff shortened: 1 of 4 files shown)


src/styles/index.css
```diff
@@ -6,4 +6,4 @@
 */
 @import '~@patternfly/patternfly/patternfly.css';
 @import '~@patternfly/patternfly/patternfly-addons.css';
-@import '~openshift-assisted-ui-lib/dist/index.css';
+@import '~openshift-assisted-ui-lib/dist/index-ocm.css';
```
jtomasek, line 9: Is this still relevant? I don't see it being produced by ui-lib

mareklibra: used to be but not anymore. We keep using `index.css`

jkilzi (changes requested):

jkilzi (approved):

PR #485, opened 2021-09-15
No-Issue: Replace fuzzysearch package with fuse.js

jgyselov:


jkilzi:
no need https://github.com/openshift-assisted/assisted-ui/pull/484

PR #503, opened 2021-11-22
Use ocm module for library imports

jtomasek:
Use dedicated ocm export for importing library modules

Depends on https://github.com/openshift-assisted/assisted-ui-lib/pull/972

### DIFF
--- a/src/components/AboutModal.tsx
+++ b/src/components/AboutModal.tsx
@@ -5,16 +5,9 @@ import {
   TextContent,
   ButtonVariant,
 } from '@patternfly/react-core';
-import { OCM } from 'openshift-assisted-ui-lib';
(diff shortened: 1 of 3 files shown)


jkilzi (dismissed):

jkilzi:
Depends on a new version of the `openshift-assisted-ui-lib` containing this change https://github.com/openshift-assisted/assisted-ui-lib/pull/972

jkilzi (approved):

jkilzi:
Unfortunately this PR won't work yet, we need to bump the TS version (btw, if we were using JS here it should have worked).
Also before we bump the TS version we need to make sure that TS version implements this feature: https://github.com/microsoft/TypeScript/issues/33079#issuecomment-878934422

PR #518, opened 2022-01-07
MGMT-8715: uses UBI images and embeds nginx config

mhrivnak:
Embedding the nginx config file eliminates the error-prone process of
making the user mount in a customized config file at runtime. The image
now takes an environment variable for the one setting that matters.

Use of UBI images brings this project closer to the rest of the
openshift ecosystem, where the entire stack is based on el8. In
particular, offering a container-based build process eliminates any
variability that can be introduced by changes in an unknown and/or
untracked external build system.

#### To do:

- [ ] test the github action that does releases
- [ ] test the k8s manifest

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,33 @@
-FROM bitnami/nginx:1.18.0
+FROM registry.access.redhat.com/ubi8/nodejs-14 as builder
+
+ENV NODE_OPTIONS=--max-old-space-size=8192
+
+ARG REACT_APP_GIT_SHA
+ENV REACT_APP_GIT_SHA=$REACT_APP_GIT_SHA
+ARG REACT_APP_VERSION
+ENV REACT_APP_VERSION=$REACT_APP_VERSION
+
+COPY --chown=1001:0 / /src/
+RUN chmod 775 /src/
+WORKDIR /src/
+
+RUN npx yarn install
+RUN npx yarn lint
+RUN npx yarn build
+
+FROM registry.access.redhat.com/ubi8/nginx-120
+
+# persist these on the final image for later inspection
+ARG REACT_APP_GIT_SHA
+ENV GIT_SHA=$REACT_APP_GIT_SHA
+ARG REACT_APP_VERSION
+ENV VERSION=$REACT_APP_VERSION
 
 COPY deploy/deploy_config.sh /deploy/
 COPY deploy/ui-deployment-template.yaml /deploy/
+COPY deploy/nginx.conf /deploy/
+COPY deploy/start.sh /deploy/
 
-COPY build/ /app/
+COPY --from=builder /src/build/ "${NGINX_APP_ROOT}/src/"
+
+CMD [ "/deploy/start.sh" ]
--- a/README.md
+++ b/README.md
@@ -76,6 +76,13 @@ Example:
 $ REACT_APP_BUILD_MODE=single-cluster yarn build
 ```
 
+## Container image build
+
+You can build the container image by running:
+```
+$ podman build -t quay.io/edge-infrastructure/assisted-installer-ui:latest . --build-arg REACT_APP_GIT_SHA="$(git rev-parse HEAD)" --build-arg REACT_APP_VERSION=latest
+```
+
 ## Available Scripts
 
 In the project directory, you can run:
--- a/deploy/nginx.conf
+++ b/deploy/nginx.conf
@@ -1,0 +1,18 @@
+index index.html;
+
+location /api {
+        proxy_pass $ASSISTED_SERVICE_URL;
+        proxy_http_version 1.1;
+        proxy_set_header Upgrade $http_upgrade;
+        proxy_set_header Connection 'upgrade';
+        proxy_set_header Host $host;
+        proxy_cache_bypass $http_upgrade;
+        proxy_connect_timeout 120;
+        proxy_send_timeout 120;
+        proxy_read_timeout 120;
+        send_timeout 120;
+}
+
+location / {
+        try_files $uri /index.html;
+}
--- a/deploy/start.sh
+++ b/deploy/start.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env sh
+
+export ASSISTED_SERVICE_URL="${ASSISTED_SERVICE_URL:-http://localhost:8090}"
+
+envsubst '$ASSISTED_SERVICE_URL' < /deploy/nginx.conf > "$NGINX_DEFAULT_CONF_PATH/nginx.conf"
+
+exec nginx -g "daemon off;"
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -9,20 +9,13 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - name: Setup Node.JS
-        uses: actions/setup-node@v2
-        with:
-          node-version: '14.x'
       - name: Set environment variables
         run: |
-          echo "NODE_OPTIONS=--max-old-space-size=8192" >> $GITHUB_ENV
           echo "REACT_APP_GIT_SHA=${{ github.sha }}" >> $GITHUB_ENV
           echo "REACT_APP_VERSION=${{ env.GIT_TAG }}" >> $GITHUB_ENV
       - run: |
-          echo NODE_OPTIONS: ${NODE_OPTIONS}
           echo REACT_APP_GIT_SHA: ${REACT_APP_GIT_SHA}
           echo REACT_APP_VERSION: ${REACT_APP_VERSION}
-          echo Node version: `node --version`
       - name: Get tag
         id: get_tag
         run: echo "GIT_TAG=${GITHUB_REF#refs/tags/}" >> $GITHUB_ENV
@@ -30,12 +23,6 @@ jobs:
           echo Detected GIT_TAG: ${GIT_TAG}
       - name: Check out code
         uses: actions/checkout@v1
-      - name: Install dependencies
-        run: yarn install
-      - name: Call linter
-        run: yarn lint
-      - name: Build code
-        run: yarn build
       - name: Publish to quay.io/edge-infrastructure
         uses: elgohr/Publish-Docker-Github-Action@3.04
         with:
@@ -45,3 +32,4 @@ jobs:
           registry: quay.io
           dockerfile: Dockerfile
           tags: "stable,${{ env.GIT_TAG }},latest-${{ github.sha }}"
+          buildargs: REACT_APP_GIT_SHA,REACT_APP_VERSION
--- a/deploy/ui-deployment-template.yaml
+++ b/deploy/ui-deployment-template.yaml
@@ -13,38 +13,6 @@ spec:
   selector:
     app: assisted-installer-ui
 ---
-kind: ConfigMap
-apiVersion: v1
-metadata:
-  name: assisted-installer-ui
-  namespace: "__NAMESPACE__"
-data:
-  nginx.conf: |
-    server {
-      listen 8080;
-      server_name _;
-
-      root /app;
-      index index.html;
-
-      location /api {
-          proxy_pass __ASSISTED_SERVICE_URL__;
-          proxy_http_version 1.1;
-          proxy_set_header Upgrade $http_upgrade;
-          proxy_set_header Connection 'upgrade';
-          proxy_set_header Host $host;
-          proxy_cache_bypass $http_upgrade;
-          proxy_connect_timeout 120;
-          proxy_send_timeout 120;
-          proxy_read_timeout 120;
-          send_timeout 120;
-      }
-
-      location / {
-         try_files $uri /index.html;
-      }
-    }
----
 kind: Deployment
 apiVersion: apps/v1
 metadata:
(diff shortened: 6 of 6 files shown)


mhrivnak:
I tested this image with the `make deploy-onprem` method in assisted-service, and it worked great.

To do:

- test the github action that does releases
- test the k8s manifest

I could use feedback on how you normally test those.

README.md
```diff
@@ -76,6 +76,13 @@ Example:
 $ REACT_APP_BUILD_MODE=single-cluster yarn build
 ```
 
+## Container image build
+
+You can build the container image by running:
```
jkilzi, line 81: @mhrivnak IMHO people my also benefit of an example showing how to override the ASSISTED_SERVICE_URL

README.md
```diff
@@ -76,6 +76,13 @@ Example:
...
+```
+$ podman build -t quay.io/edge-infrastructure/assisted-ui:latest . --build-arg REACT_APP_GIT_SHA="$(git rev-parse HEAD)" --build-arg REACT_APP_VERSION=latest
+```
+
```
jkilzi, line 85: @mhrivnak IMHO people may also benefit from an example showing how to override the ASSISTED_SERVICE_URL

deploy/start.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env sh
+
+export ASSISTED_SERVICE_URL="${ASSISTED_SERVICE_URL:-http://localhost:8090}"
```
jkilzi, line 3: If the ASSISTED_SERVICE_URL env-var isn't persisted in the Dockerfile. Would we be able to override it?

Dockerfile
```diff
@@ -1,6 +1,33 @@
...
+RUN chmod 775 /src/
+WORKDIR /src/
+
+RUN npx yarn install
+RUN npx yarn lint
+RUN npx yarn build
+
+FROM registry.access.redhat.com/ubi8/nginx-120
+
+# persist these on the final image for later inspection
```
jkilzi, line 20: Don't we need this too?
```suggestion
# persist these on the final image for later inspection
ARG ASSISTED_SERVICE_URL
ENV ARG ASSISTED_SERVICE_URL=$ASSISTED_SERVICE_URL
```

Dockerfile
```diff
@@ -1,6 +1,33 @@
...
 COPY deploy/deploy_config.sh /deploy/
 COPY deploy/ui-deployment-template.yaml /deploy/
+COPY deploy/nginx.conf /deploy/
+COPY deploy/start.sh /deploy/
+
+COPY --from=builder /src/build/ "${NGINX_APP_ROOT}/src/"
 
-COPY build/ /app/
\ No newline at end of file
+CMD [ "/deploy/start.sh" ]
```
jkilzi, line 33: Following [this guie](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#entrypoint), wouldn't it be better to use the ENTRYPOINT directive in this case, WDYT?
```suggestion
ENTRYPOINT [ "/deploy/start.sh" ]
```

Dockerfile
```diff
@@ -1,6 +1,33 @@
...
+ARG REACT_APP_VERSION
+ENV REACT_APP_VERSION=$REACT_APP_VERSION
+
+COPY --chown=1001:0 / /src/
+RUN chmod 775 /src/
+WORKDIR /src/
+
+RUN npx yarn install
+RUN npx yarn lint
+RUN npx yarn build
```
jkilzi, line 16: Perfect!

deploy/nginx.conf
```diff
@@ -0,0 +1,24 @@
+server {
+  listen 8080;
+  server_name localhost;
+
+  root $NGINX_APP_ROOT/src/;
+  index index.html;
+
+  location /api {
+      proxy_pass $ASSISTED_SERVICE_URL;
```
jkilzi, line 9: The value here may differ from the value in the deployment template, is that OK?

Dockerfile
jkilzi, line 16: @mhrivnak after taking your branch for a test-drive, and despite you managed to build the app correctly, I'll advise against building the app in a multi-stage fashion for a few reasons.

1. It slows down the build process. 2 images need to be pulled for each release run.
2. Parsing the output log from each step (install/lint/build) is not as transparent as it is while building inside the build machine itself.
3. The build environment is less important because we won't be using node.js after the `build` folder is created. What matters eventually is to have the `build` folder embedded in the container and served via `nginx`.

#### I suggest having something like this:
```
FROM registry.access.redhat.com/ubi8/nginx-120

COPY --chown=1001:0 /build/ "${NGINX_APP_ROOT}/src/"

# persist these on the final image for later inspection
ARG REACT_APP_GIT_SHA
ENV GIT_SHA=$REACT_APP_GIT_SHA
ARG REACT_APP_VERSION
ENV VERSION=$REACT_APP_VERSION

WORKDIR /deploy/
COPY deploy/deploy_config.sh /deploy/
COPY deploy/ui-deployment-template.yaml /deploy/
COPY deploy/nginx.conf /deploy/
COPY deploy/start.sh /deploy/


ENTRYPOINT [ "/deploy/start.sh" ]

```

README.md
```diff
@@ -76,6 +76,13 @@ Example:
 $ REACT_APP_BUILD_MODE=single-cluster yarn build
 ```
 
+## Container image build
+
+You can build the container image by running:
+```
+$ podman build -t quay.io/edge-infrastructure/assisted-ui:latest . --build-arg REACT_APP_GIT_SHA="$(git rev-parse HEAD)" --build-arg REACT_APP_VERSION=latest
```
jkilzi, line 83:
```suggestion
$ podman build -t quay.io/edge-infrastructure/assisted-installer-ui:latest . --build-arg REACT_APP_GIT_SHA="$(git rev-parse HEAD)" --build-arg REACT_APP_VERSION=latest
```

jkilzi (changes requested):

jkilzi:
> I tested this image with the `make deploy-onprem` method in assisted-service, and it worked great.
> 
> To do:
> 
> * test the github action that does releases
> * test the k8s manifest
> 
> I could use feedback on how you normally test those.
> 
> And by the way: everything I know about building a react app and using yarn I learned in the past 24 hours just to accomplish this, so please give my work an extra-thorough review!

In order to test the k8s manifest you can simply build the image locally, generate the manifest and then apply it to a running k8s instance (e.g. Minikube).

#### How to generate the k8s manifest:
```
podman run --rm -it --entrypoint /deploy/deploy_config.sh quay.io/edge-infrastructure/assisted-ui -t /deploy/ui-deployment-template.yaml
```

##### `/deploy/deploy_config.sh` options:
```
podman run --rm -it --entrypoint /deploy/deploy_config.sh quay.io/edge-infrastructure/assisted-ui -h

Usage: /deploy/deploy_config.sh [ -t deploy-config-template ] [ -u ASSISTED-SERVICE-URL ] [ -i UI_IMAGE_NAME ] [ -n NAMESPACE ]
```

Testing the release GH action requires running the workflow for which a new tag needs to be created. We can figure out how to do this in a controlled way later.

Dockerfile
mhrivnak: It depends. ENTRYPOINT is useful when you expect there to be arguments that a user could/should supply at runtime. The downside is that you cannot override the ENTRYPOINT at runtime. It is sometimes useful for development, testing, or troubleshooting to run different commands than the one specified in the Dockerfile, and this option prevents that.

Using just CMD gives you the option to specify a different command at runtime. Many images take that approach, including the UBI nginx base image as a relevant example.

It's not likely to make a big difference either way, but I would retain the flexibility of CMD unless there was something about the ENTRYPOINT we definitely want to take advantage of.

Dockerfile
jkilzi: This includes reverting these steps in the release.yaml file.

mhrivnak: 1. I assume devs would still want to build on their local machine. But for releases, does a few extra minutes make a difference? Some projects have multiple Dockerfiles for this reason, so maybe it would be a reasonable compromise to have a `Dockerfile.dev` that uses the `build` dir copied in as today?

2. Can you elaborate? How is it different? I would think that whether it runs inside or outside a container, we should be able to get the same output.

3. The main benefit here of a multistage build is having a reference build environment that we can inspect and control, and is the same no matter where it runs. You have the same exact build environment (and results) when using github actions, your laptop, someone else's laptop, an automated quay.io build, openshift CI, or anywhere else. It's valuable for repeatability/reproducibility, and a "software supply chain" perspective (build tools compiled by Red Hat, vs. by ??? [See page 42 here for more](https://github.com/cncf/tag-security/blob/main/supply-chain-security/supply-chain-security-paper/CNCF_SSCP_v1.pdf)). AFAIK all openshift images that build from source are built using multi-stage builds for this reason. You can also see the other assisted-installer containers get built this way. Of course this image is not currently part of a product, but it is being used by the openshift community, and I think we're wise to follow similar best practices.

From my own perspective, the multi-stage build enabled me to build the project from source without installing any dependencies on my computer. That was helpful, and why I made the change as part of this PR

If this is a blocking issue, I can move the proposal for a multi-stage build to a separate PR. But I do think it's a significant best practice worth following, and it enabled me to build the image without having to worry about setting up a build environment myself.

jkilzi: > 1. I assume devs would still want to build on their local machine. But for releases, do a few extra minutes make a difference? Some projects have multiple Dockerfiles for this reason, so maybe it would be a reasonable compromise to have a `Dockerfile.dev` that uses the `build` dir copied in as today?  

"Do a few extra minutes make a difference?" - Yes, especially If the process can take just a few seconds.
"I assume devs would still want to build on their local machine" - Depends on their motivation...
If their motivation is to contribute to this project then they will need to set up their own dev environment anyway (even if that is inside a container). Anyway, offering a container development environment is an interesting and useful topic that I'll be happy to explore but in a different thread.

> 2. Can you elaborate? How is it different? I would think that whether it runs inside or outside a container, we should be able to get the same output.  

Sure. Investigating log failures (manually) is never a pleasant task. For that reason, GH actions parse the output of each build step in a clear way. You can tell easily what output belongs to which step, collapse sections that are not of interest, or expand a particular step section after it has failed to ease the task of troubleshooting the issue. On the other hand, having the build machine just call one build step (docker/podman build...) will defeat the purpose of this feature, you'll end up having a very large output that is not split into sections and makes it harder to investigate.
Since building in the build machine provided by GH and inside a container does produce the same output, I tend to favor the first approach. BTW, I would favor the container approach if I would have to use a shared build machine, that provides no guaranty on whether it has been cleaned up properly after its last use or not. This is not the case.

> 3. The main benefit here of a multistage build is having a reference build environment that we can inspect and control, and is the same no matter where it runs. You have the same exact build environment (and results) when using github actions, your laptop, someone else's laptop, an automated quay.io build, openshift CI, or anywhere else. It's valuable for repeatability/reproducibility, and a "software supply chain" perspective (build tools compiled by Red Hat, vs. by ??? [See page 42 here for more](https://github.com/cncf/tag-security/blob/main/supply-chain-security/supply-chain-security-paper/CNCF_SSCP_v1.pdf)). AFAIK all openshift images that build from source are built using multi-stage builds for this reason. You can also see the other assisted-installer containers get built this way. Of course, this image is not currently part of a product, but it is being used by the openshift community, and I think we're wise to follow similar best practices.  

I understand and second your perspective here. Security-wise I have no objection. Just bear in mind that the first action performed in the release pipeline, the one that checks out the code, is performed by a tool (GH predefined [actions/checkout](https://github.com/actions/checkout)) that is not reviewed by Red Hat. If software sources other than Red Hat's shall not be trusted then this can still manifest into an attack vector in a supply chain attack. Now having considered this scenario, why didn't you rather check out the sources from within the UBI? Which software sources can be trusted and which not? How many of the build tasks can be securely delegated to the build machine provided by GH? If the answer is none, then there is no discussion and we should absolutely do a multi-stage build since there's no other option.

> From my own perspective, the multi-stage build enabled me to build the project from source without installing any dependencies on my computer. That was helpful, and why I made the change as part of this PR

You can still achieve this. Just pull the node.js UBI locally, map a directory between the container and your local filesystem in order to gain access to the build directory after it is created and then perform the exact 3 steps you did in the builder stage.

mhrivnak: Ok, it sounds like the nice interface of github actions is a big motivator for continuing to use it.

For releases, I still don't understand; it happens once every few days and is automated. Why does every minute or second count?

As for the supply chain aspect, yes of course using a build container does not solve all of the problems. However it is a big step forward in reproducibility. Anyone can easily reproduce the build environment without depending on github's CI.

All that said, we could discuss this for a long time. :) How would you like to move toward a decision? Talk with the team? Save this part for later? Shoot it down for good?

jkilzi: Let me take it with the team. Tomorrow we'll settle the discussion.

mareklibra:
Practical-wise, I do not see any difference for our so-far use-cases, so I am ok with merging it.

@mhrivnak, can we shortly discuss (probably offline) the motivation behind that change? Is it expected to be reused in the downstream build? Any news about shipping it? What other runners will execute that build? Etc.

jtomasek (dismissed):

Dockerfile
jkilzi: ack, 10x for the explanation

jkilzi (dismissed):

mhrivnak:
The PR is ready to go from my perspective and works both when deployed locally in podman and in a cluster. The only part remaining is to test the github action, which I'll need your help accomplishing.

I tested this in-cluster and was able to further slim down the required nginx config. I deployed assisted-service with `make deploy-all TARGET=oc-ingress` (which itself had some issues I had to fix to get it working). I then rendered the template, applied the resulting manifest, and created a Route for the UI's service. I set `ASSISTED_SERVICE_URL=http://assisted-service:8090`, so it would resolve directly to the assisted-service's Service.

mareklibra:
Starting the CI fot that

mareklibra:
CI passed

mareklibra:
@mhrivnak What about updating the remaining two github workflows to use the containerized build as well? To stay aligned.

jkilzi:
> @mhrivnak What about updating the remaining two github workflows to use the containerized build as well? To stay aligned.

I'll take care of that

jkilzi (approved):

PR #538, opened 2022-02-22
Fixed the Assisted UI not working with disabled IPv6 issue

lveyde:
nginx log the following error: `nginx: [emerg] socket() [::]:8080 failed (97: Address family not supported by protocol)`

Due to the known issue with the Nginx, disabling IPv6 in the default configuration will cause it to fail to start.

This patch fixes this issue by testing during the runtime for disabled IPv6 and updating the Nginx configuration.

https://issues.redhat.com/browse/MGMT-9342

### DIFF
--- a/deploy/start.sh
+++ b/deploy/start.sh
@@ -4,4 +4,10 @@ export ASSISTED_SERVICE_URL="${ASSISTED_SERVICE_URL:-http://localhost:8090}"
 
 envsubst '$ASSISTED_SERVICE_URL' < /deploy/nginx.conf > "$NGINX_DEFAULT_CONF_PATH/nginx.conf"
 
+# Do not listen on IPv6 if it's not enabled in the hardware
+if grep 'ipv6.disable=1' /proc/cmdline; then
+    cp "${NGINX_CONF_PATH}" ~/nginx.conf.orig
+    sed -e 's/listen\s*\[::\].*//g' ~/nginx.conf.orig | dd of="${NGINX_CONF_PATH}"
+fi
+
 exec nginx -g "daemon off;"


jkilzi:
@lveyde can you please add a reference/source to this known issue related to IPv6 that affects the version of the ngnix image we are using in order to build the image?

jkilzi:
The closest issue I could find is this one: https://access.redhat.com/solutions/4898101
The error indicates _"that ipv6 is disabled on the server, hence service is and failed due to unsupported address family."_

deploy/start.sh
```diff
@@ -4,4 +4,9 @@ export ASSISTED_SERVICE_URL="${ASSISTED_SERVICE_URL:-http://localhost:8090}"
 
 envsubst '$ASSISTED_SERVICE_URL' < /deploy/nginx.conf > "$NGINX_DEFAULT_CONF_PATH/nginx.conf"
 
+if grep 'ipv6.disable=1' /proc/cmdline; then
```
jkilzi, line 7:
```suggestion
# Do not listen on IPv6 if it's not enabled in the hardware
if grep 'ipv6.disable=1' /proc/cmdline; then
```

jkilzi (approved):

lveyde:
> 

The issue seems to be known as it affects a lot of Nginx users i.e. :

https://bugs.launchpad.net/ubuntu/+source/nginx/+bug/1743592
https://www.ibm.com/docs/it/aspera-shares/1.9?topic=a-disabling-ipv6-support-in-shares
https://github.com/mail-in-a-box/mailinabox/issues/1783

PR #573, opened 2022-06-29
MGMT-8156: Extract strings for localization

ammont82:
To resolve problems with localization implemented in assisted-ui-lib we need to add the i18n configuration to assisted-ui project.

Related to https://github.com/stolostron/console/pull/1710

Related to https://issues.redhat.com/browse/MGMT-8156

More info: https://docs.google.com/document/d/1KxjA8qsKA3K81MQrMF4jt0Ope4EM0lHWAN0emUx-q4w/edit

Depends on https://github.com/openshift-assisted/assisted-ui-lib/pull/1484.


src/i18n.js
```diff
@@ -0,0 +1,49 @@
...
+  .use(LanguageDetector)
+  // init i18next
+  // for all options read: https://www.i18next.com/overview/configuration-options
+  .init({
+    resources,
+    supportedLngs: ['en'],
+    fallbackLng: 'en',
+    load: 'languageOnly',
+    detection: { caches: [] },
+    contextSeparator: '~',
```
jtomasek, line 33: We should not override this. Keep it default ('_')

src/i18n.js
```diff
@@ -0,0 +1,49 @@
+/* Copyright Contributors to the Open Cluster Management project */
+
+/* istanbul ignore file */
+
+import i18n from 'i18next';
+import { initReactI18next } from 'react-i18next';
+import LanguageDetector from 'i18next-browser-languagedetector';
+import translation from './locales/en/translation.json';
```
jtomasek, line 8: Don't copy the file to assisted-ui. The translations should always be imported from assisted-ui-lib. They are available from
```
import locales from 'openshift-assisted-ui-lib/locales'
```

src/i18n.js
```diff
@@ -0,0 +1,49 @@
...
+/** Localization */
+// the translations
+// the translations
+const resources = {
+  en: {
+    /* eslint-disable @typescript-eslint/camelcase */
+    assisted_installer: translation,
```
jtomasek, line 15: this namespace is incorrect IIUC

src/i18n.js
```diff
@@ -0,0 +1,49 @@
...
+  },
+};
```
jtomasek, line 17:
```suggestion
const resources = locales;
```

jtomasek (changes requested):

celdrake:
Please, update also the README to reflect the new environment variable and indicate which value should be used.

celdrake (approved):

jkilzi (approved):

PR #600, opened 2022-09-01
MGMT-11857: Support nginx SSL configuration

nmagnezi:
In a case where ASSISTED_SERVICE_SCHEME is set to https, the
generated nginx configuration should be set to ssl

### DIFF
--- a/deploy/nginx_ssl.conf
+++ b/deploy/nginx_ssl.conf
@@ -1,0 +1,9 @@
+ server {
+    listen 443 ssl;
+
+    ssl_certificate     $HTTPS_CERT_FILE;
+    ssl_certificate_key $HTTPS_KEY_FILE;
+
+    ssl_verify_client       on;
+    ssl_ocsp                on;
+}
--- a/deploy/start.sh
+++ b/deploy/start.sh
@@ -4,6 +4,12 @@ export ASSISTED_SERVICE_URL="${ASSISTED_SERVICE_URL:-http://localhost:8090}"
 
 envsubst '$ASSISTED_SERVICE_URL' < /deploy/nginx.conf > "$NGINX_DEFAULT_CONF_PATH/nginx.conf"
 
+if [ "$ASSISTED_SERVICE_SCHEME" = "https" ]; then
+    envsubst 'HTTPS_CERT_FILE' < /deploy/nginx_ssl.conf > "/deploy/nginx_ssl.conf.tmp"
+    envsubst 'HTTPS_KEY_FILE'  < /deploy/nginx_ssl.conf.tmp > "$NGINX_DEFAULT_CONF_PATH/nginx_ssl.conf"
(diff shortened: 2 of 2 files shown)


nmagnezi:
/hold
testing

deploy/nginx_ssl.conf
```diff
@@ -0,0 +1,12 @@
...
+    listen 443 ssl;
+
+    ssl_certificate     $HTTPS_CERT_FILE;
+    ssl_certificate_key $HTTPS_KEY_FILE;
+
+    ssl_verify_client       on;
+    ssl_trusted_certificate /etc/ssl/cachain.pem;
+    ssl_ocsp                on; # Enable OCSP validation
+
+    #...
```
jkilzi, line 11:
```suggestion
    #...
```
What's this? :)

nmagnezi: removed :)

deploy/start.sh
```diff
@@ -4,6 +4,12 @@ export ASSISTED_SERVICE_URL="${ASSISTED_SERVICE_URL:-http://localhost:8090}"
 
 envsubst '$ASSISTED_SERVICE_URL' < /deploy/nginx.conf > "$NGINX_DEFAULT_CONF_PATH/nginx.conf"
 
+if [ "$ASSISTED_SERVICE_SCHEME" = "https" ]; then
+    envsubst 'HTTPS_CERT_FILE' < /deploy/nginx_ssl.conf > "/deploy/nginx_ssl.conf.tmp"
+    envsubst 'HTTPS_KEY_FILE'  < /deploy/nginx_ssl.conf.tmp > "$NGINX_DEFAULT_CONF_PATH/nginx_ssl.conf"
```
jkilzi, line 9:
```suggestion
    envsubst 'HTTPS_KEY_FILE'  < /deploy/nginx_ssl.conf.tmp > "$NGINX_DEFAULT_CONF_PATH/nginx_ssl.conf"
```
How does ngnix know it should use this file instead of `nginx.conf`. We'll end up having both files inside `$NGINX_DEFAULT_CONF_PATH/` when `"$ASSISTED_SERVICE_SCHEME" = "https"`

jkilzi (changes requested):

deploy/start.sh
nmagnezi: See this `nginx.conf` (grabbed it from the UI container): https://gist.github.com/nmagnezi/58a7e6e2f87fc1ff820aaea656738927#file-nginx-conf-L65-L66

Lines 65-66:
```
#        # Load configuration files for the default server block.
#        include /opt/app-root/etc/nginx.default.d/*.conf;
```

jkilzi (approved):

nmagnezi:
@jkilzi please revert. I'm still testing and we should not have merged this as-is.
I tried to place a hold label.. seems like it does not work here

PR #605, opened 2022-09-09
Change axios version to fix problems with polling

ammont82:
We need to change the axios version to be >=0.22 to use AbortController in assisted-ui-lib project.
Related to https://github.com/openshift-assisted/assisted-ui-lib/pull/1628


package.json
```diff
@@ -16,7 +16,7 @@
     "@patternfly/react-tokens": "^4.58.2",
     "@reduxjs/toolkit": "~1.6.x",
     "@sentry/browser": "5.19.1",
-    "axios": "^0.21.2",
+    "axios": "^0.22.0",
```
jkilzi, line 19:
```suggestion
    "axios": ">=0.22.0 <1.0.0",
```

Usage of caret for packages starting with a 0-major version is problematic. It only updates the patch version!

jkilzi (changes requested):

package.json
ammont82: Ok!

celdrake (approved):

PR #606, opened 2022-09-18
NO-ISSUE: fixes git-sha and lib version display values

jkilzi:
This PR fixes the values displayed (in dev-mode) by the env. vars REACT_APP_VERSION and REACT_APP_GIT_SHA.  

- Removes old scripts that are no longer in use from the hacks folder
- REACT_APP_VERSION now displays the version of the ui-lib located inside the node_modules
- REACT_APP_GIT_SHA show the assisted-ui's current git-sha
- REACT_APP_API_URL needs to be defined inside a .env.development.local file (not kept under VCS, so every dev can have their own settings)


ammont82 (approved):

hacks/setup-env.sh
```diff
@@ -0,0 +1,7 @@
+REACT_APP_API_URL="http://REACT_APP_API_URL.env.variable.not:set"
```
celdrake, line 1: Shouldn't the script include the `#!/bin/bash` directive?

hacks/setup-env.sh
```diff
@@ -0,0 +1,7 @@
...
+REACT_APP_VERSION="$(node -e 'console.log(require("./node_modules/openshift-assisted-ui-lib/package.json").version)')"
```
celdrake, line 2: Maybe this a faster alternative?

`cat node_modules/openshift-assisted-ui-lib/package.json | jq -r '.version'`

celdrake:
@jkilzi We should update the file `create-test.mjs` as it is being referenced in the initial comment with its previous name `setup-env.mjs`

hacks/setup-env.sh
jkilzi: no because this file is meant to be sourced, no shell will execute the code

hacks/setup-env.sh
jkilzi: Yes, but that assumes that you are using Linux and have `jq` installed.
No matter what OS you will use in order to execute the project, you must have node installed there.

hacks/setup-env.sh
celdrake: That's true, thanks

hacks/setup-env.sh
celdrake: True also, specially `jq` would not be available in most cases.

celdrake (approved):

PR #627, opened 2023-01-02
Single cluster mode deprecation

jkilzi:
Follows up on https://github.com/openshift-assisted/assisted-ui-lib/pull/1829

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -14,7 +14,6 @@ const {
   Store: { store },
   Router: LibRouter,
   Features,
-  Config,
 } = OCM;
 
 const App: React.FC = () => (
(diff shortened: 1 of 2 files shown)


src/components/App.tsx
```diff
@@ -28,13 +27,7 @@ const App: React.FC = () => (
...
         defaultManagedSidebarIsOpen={false}
       >
-        <LibRouter
-          features={
-            Config.isSingleClusterMode()
-              ? Features.SINGLE_CLUSTER_ENABLED_FEATURES
-              : Features.STANDALONE_DEPLOYMENT_ENABLED_FEATURES
-          }
-        />
+        <LibRouter features={Features.STANDALONE_DEPLOYMENT_ENABLED_FEATURES} />
```
celdrake, line 30: We should also remove `SINGLE_CLUSTER_ENABLED_FEATURES`

It doesn't seem to be referenced in UHC Portal.

celdrake, line 30: We should remove the function `isSingleClusterMode`

celdrake (changes requested):

src/components/App.tsx
celdrake: I see, these come from Assisted-ui-lib. All good.

celdrake: same, it's defined in Assisted UI lib.

celdrake (approved):

